PR #90, opened 2022-07-09
Remove looper from clientlib

sdlaver:


android/fakedapp/src/main/java/com/solana/mobilewalletadapter/fakedapp/MainViewModel.kt
```diff
@@ -19,20 +18,20 @@ import kotlinx.coroutines.*
...
 import java.util.concurrent.TimeoutException
 import kotlin.random.Random
 
-class MainViewModel(application: Application) : AndroidViewModel(application) {
+class MainViewModel : ViewModel() {
     private val _uiState = MutableStateFlow(UiState())
     val uiState = _uiState.asStateFlow()
 
-    private val mobileWalletAdapterClientMutex = Mutex() // allow only a single MWA connection at a time
+    private val mobileWalletAdapterClientSem = Semaphore(1) // allow only a single MWA connection at a time
```
sdlaver, line 34: I played with a few other options here - using a channel, using a special coroutine scope with a known single-threaded context, and a Semaphore. Ultimately, I think that the Semaphore is the appropriate abstraction. The concept of permits better matches what we are trying to express than a Mutex (which implies a concrete resource).

creativedrewy: So you may not be wrong, but again, this seems to be overkill. An easier way would be to just enforce this at the UI level. Have the ViewModel ViewState have an "isWorking" property that gets set to true at the beginning of each operation and set to false when done. The UI could check that value before kicking off any operation.

I guess what I'm thinking through is what happens if the user taps two buttons in rapid succession? Does the semaphore just keep the second operation from starting? Does it "magically" queue the second op? The semaphore/Mutex actually makes it unclear what actually will happen. Enforcing at the ViewState/UI level is clear about what will and won't happen.

creativedrewy (commented):
Moving in the right direction!

android/clientlib/src/main/java/com/solana/mobilewalletadapter/clientlib/scenario/LocalAssociationScenario.java
```diff
@@ -109,85 +110,98 @@ public Intent createAssociationIntent() {
...
+                default:
+                    throw new IllegalStateException("Error: attempt to close in an unknown state");
+            }
         }
 
         return future;
     }
 
+    @NonNull
+    @GuardedBy("mLock")
```
creativedrewy, line 186: What does this achieve from an operational perspective?

creativedrewy: Also, in case you hadn't noticed, I'm not too familiar with advanced operations relating to the JVM's threading model. But that's because I haven't had to for the most part. 99% of the time in my Android app experience, relatively simple thread management is enough, as long as I'm keeping the Android lifecycle in mind. I just want to make sure we are enabling as simple operations as we can for our consumers.

creativedrewy: Final point: I'm not saying this is wrong. I just wanna make sure my simpleton brain can understand this stuff. If so, then we are probably on the right track 😁

android/fakedapp/src/main/java/com/solana/mobilewalletadapter/fakedapp/MainViewModel.kt
sdlaver: I envision this as a situation that will probably be abstracted away by the convenience wrappers, which are better positioned to have an opinion on how multiple consumers will be handled.

In terms of the technical details, it acts like a regular Semaphore would, but substitute suspending for blocking. For e.g., if a second consumer attempts to acquire a permit when none are available, the coroutine context will be suspended until one is available. Since it's a coroutine sync primitive, it respects cancelability.

android/clientlib/src/main/java/com/solana/mobilewalletadapter/clientlib/scenario/LocalAssociationScenario.java
sdlaver: The `@GuardedBy` annotation is for static analysis tools - it informs the tool what lock should be held when this method is called. It gives hints to call flow analysis tools to catch calls to methods without holding a required lock.

This is purely an implementation detail of LocalAssociationScenario; with this PR, none of the threading details leak out through the public APIs.

android/fakedapp/src/main/java/com/solana/mobilewalletadapter/fakedapp/MainViewModel.kt
creativedrewy: Again, my main concern is the type of code we're putting out there as the "right" way to do things. I don't necessarily want people to think they have to do complicated things just to get baseline functionality.

That said, if you're cool with the fake dApp being rewritten to use one of the convenience wrappers once we create them, I'm cool with this.

olivierg13: Yeah I agree:
- Complex code is often needed to handle complex problems and situations. This code looks good to me, funnily it reminds me exactly of the logic for "Downloads" which has similar problems to solve I guess.
- Hiding this complex code behind easier-to-use or easier-to-understand APIs might be the next priority, also by making these new APIs the default implementation in the example apps (fake DApp etc...).

I guess one question I would have would be: in addition to the Coroutines and Rx APIs, are we thinking about having a lightweight (I.E. without important external dependencies) wrapper that simplifies the implementation for devs? Like using an Executor, or something like that.

creativedrewy (approved):
Good to go!

PR #96, opened 2022-07-11
Add clientlib-ktx module files

creativedrewy:
This is _just_ the module files. Impl is next.


android/clientlib-ktx/proguard-rules.pro
```diff
@@ -0,0 +1,21 @@
+# Add project specific ProGuard rules here.
```
olivierg13, line 1: Nitpick: the 2 proguard files are probably not needed?

android/settings.gradle
```diff
@@ -22,3 +22,4 @@ include ':walletlib'
 include ':clientlib'
 include ':fakewallet'
 include ':fakedapp'
+include ':clientlib-ktx'
```
olivierg13, line 25: Nitpick++ 😁 : you could put it just under `include ':clientlib'`

olivierg13 (approved):

android/settings.gradle
creativedrewy: No, it's good to keep this organized!

android/clientlib-ktx/proguard-rules.pro
olivierg13: Good question 🤔 Usually proguarding libraries (especially open-source) is not advised since it would make it more difficult to debug/reverse for developers importing it. I guess we could just run the "optimization" part of proguard, that's probably not a priority?

creativedrewy: I'll just remove for now; if we need it we can re-add.

android/clientlib-ktx/build.gradle
```diff
@@ -0,0 +1,42 @@
+plugins {
+    id 'com.android.library'
+    id 'org.jetbrains.kotlin.android'
+}
+
+android {
+    compileSdk 32
+
+    defaultConfig {
+        minSdk 19
```
sdlaver, line 10: This is less than `clientlib` - let's standardize on `23`

sdlaver (approved):

android/clientlib-ktx/build.gradle
creativedrewy: You won't get an argument from me!

PR #99, opened 2022-07-12
Adding Clientlib RxJava APIs

olivierg13:


android/clientlib-rxjava/src/main/java/com/solana/mobilewalletadapter/clientlib/protocol/RxMobileWalletAdapterClient.java
```diff
@@ -0,0 +1,34 @@
...
+    public RxMobileWalletAdapterClient(@IntRange(from = 0) int clientTimeoutMs) {
+        mMobileWalletAdapterClient = new MobileWalletAdapterClient(clientTimeoutMs);
+    }
+
+    @NonNull
+    public Single<AuthorizeResult> authorize(@Nullable Uri identityUri,
+                                             @Nullable Uri iconUri,
+                                             @Nullable String identityName) {
+        try {
+            AuthorizeFuture authorizeFuture = mMobileWalletAdapterClient.authorize(identityUri, iconUri, identityName);
```
creativedrewy, line 28: How are you thinking about handling association creation/starting/closing? Is that going to be a different containing method?

olivierg13: Do you mean letting people know that they need to subscribe to these Observables to "launch" the underlying Future?

We could use the `@CheckResult` annotation to make sure of that 🤔 , let me do that.

Or that wasn't the actual context you were describing haha 😁

PR #101, opened 2022-07-12
Adding Kotlin convenience MobileWalletAdapter APIs

creativedrewy:


android/clientlib-ktx/src/main/java/com/solana/mobilewalletadapter/clientlib/TransactionParams.kt
```diff
@@ -0,0 +1,25 @@
...
+
+open class TransactionParams(
+    val commitmentLevel: CommitmentLevel,
+    val cluster: RpcCluster,
+    val skipPreflight: Boolean,
+    val preflightCommitment: CommitmentLevel?
+)
+
+//TODO: We can add other defaults as they become relevant
+object DefaultTestnet : TransactionParams(
```
creativedrewy, line 20: Thought it might be nice to have some "default" options people could use.

android/clientlib-ktx/src/main/java/com/solana/mobilewalletadapter/clientlib/scenario/ScenarioChooser.kt
```diff
@@ -0,0 +1,14 @@
+package com.solana.mobilewalletadapter.clientlib.scenario
+
+class ScenarioChooser {
```
creativedrewy, line 3: The idea here is to have a class that can find and create the right scenario class based on the given type as other options are coming in the future. Thoguhts?

creativedrewy (commented):
Some comments from me about implementation.

android/clientlib-ktx/src/main/java/com/solana/mobilewalletadapter/clientlib/scenario/ScenarioChooser.kt
sdlaver: There are pretty different actions that need to be taken depending on the chosen association method. For example, displaying a QR code, sending an Intent, etc. It seems that choosing the scenario is already a fairly involved operation, and this helper class may not actually end up abstracting away enough of the details to be useful? Or would the intention be to have this class act as a pretty complex factory that can produce everything that is required by a given Scenario?

android/clientlib-ktx/src/main/java/com/solana/mobilewalletadapter/clientlib/TransactionParams.kt
sdlaver: The protocol itself defines defaults. Perhaps the right answer here is to allow these to be nullable, and defer to protocol (or wallet) defaults if not specified?

android/clientlib-ktx/src/main/java/com/solana/mobilewalletadapter/clientlib/MobileWalletAdapter.kt
```diff
@@ -0,0 +1,187 @@
...
+import kotlinx.coroutines.CoroutineDispatcher
+import kotlinx.coroutines.Dispatchers
+import kotlinx.coroutines.withContext
+import java.io.IOException
+import java.util.concurrent.CancellationException
+import java.util.concurrent.ExecutionException
+import java.util.concurrent.TimeUnit
+import java.util.concurrent.TimeoutException
+
+fun LocalAssociationScenario.begin(): MobileWalletAdapterClient {
```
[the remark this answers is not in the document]
sdlaver: Yep, SGTM

android/clientlib-ktx/src/main/java/com/solana/mobilewalletadapter/clientlib/MobileWalletAdapter.kt
```diff
@@ -0,0 +1,187 @@
...
+    return start().get(MobileWalletAdapter.ASSOCIATION_TIMEOUT_MS, TimeUnit.MILLISECONDS)
```
sdlaver, line 23: There are some legitimate exceptions that will be thrown by `start()` that should either be handled here (and made friendly), or documented as thrown by this method. For example, if the server drops the connection for any reason before the session is fully established, this will throw a `LocalAssociationScenario.ConnectionFailedException`

android/clientlib-ktx/src/main/java/com/solana/mobilewalletadapter/clientlib/MobileWalletAdapter.kt
```diff
@@ -0,0 +1,187 @@
...
+}
+
+fun LocalAssociationScenario.end() {
+    close().get(MobileWalletAdapter.ASSOCIATION_TIMEOUT_MS, TimeUnit.MILLISECONDS)
+}
+
+@Suppress("BlockingMethodInNonBlockingContext")
```
sdlaver, line 30: My preference is always to suppress at the smallest scope possible. Can this be applied to the statement or function instead?

android/clientlib-ktx/src/main/java/com/solana/mobilewalletadapter/clientlib/MobileWalletAdapter.kt
```diff
@@ -0,0 +1,187 @@
...
+    private val timeout: Int = Scenario.DEFAULT_CLIENT_TIMEOUT_MS,
+    private val ioDispatcher: CoroutineDispatcher = Dispatchers.IO
+) {
+    private val contract = LocalAssociationIntentContract()
+    private val resultLauncher = resultCaller.registerForActivityResult(contract) { }
+
+    private val scenarioChooser = ScenarioChooser()
+
+    suspend fun authorize(identityUri: Uri, iconUri: Uri, identityName: String): Boolean {
+        return withScenario(associate()) { client ->
```
sdlaver, line 43: Here and below - if you want to call several MWA methods back-to-back, this is going to continuously switch to and return from the wallet. In addition to the one-off methods defined ehre, what about either a Builder object, to create a chain of requests, or a closure provided to a `run` method that allows chaining together multiple calls?

android/clientlib-ktx/build.gradle
```diff
@@ -29,11 +29,16 @@ android {
     kotlinOptions {
         jvmTarget = '1.8'
     }
+
+    buildFeatures {
+        buildConfig = false
```
sdlaver, line 34: Can you detail why this was needed?

creativedrewy: It was trying to generate BuildConfig, which was colliding with the core library one.

android/clientlib-ktx/src/main/java/com/solana/mobilewalletadapter/clientlib/MobileWalletAdapter.kt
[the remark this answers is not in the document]
creativedrewy: Will do!

android/clientlib-ktx/src/main/java/com/solana/mobilewalletadapter/clientlib/MobileWalletAdapter.kt
creativedrewy: Did you see my `withScenario` method? I could move the start-catching exceptions to here too, to make it clearer.

android/clientlib-ktx/src/main/java/com/solana/mobilewalletadapter/clientlib/MobileWalletAdapter.kt
creativedrewy: Yes it can. Will do.

android/clientlib-ktx/src/main/java/com/solana/mobilewalletadapter/clientlib/MobileWalletAdapter.kt
creativedrewy: Yes, I was actually thinking about that! I will do that as part of this PR. With a few small changes I think I can make something nice.

android/clientlib-ktx/src/main/java/com/solana/mobilewalletadapter/clientlib/TransactionParams.kt
creativedrewy: Well if you want to set custom values, you can just instantiate the `TransactionParams` class from line 12. Both `CommittmentLevel` values should be nullable though?

android/clientlib-ktx/src/main/java/com/solana/mobilewalletadapter/clientlib/scenario/ScenarioChooser.kt
creativedrewy: I was wondering about this. It's probably too much to ask to know ahead of time what we're going to need from this class. This class already isn't that great, so it's just going to get in our way. I'll remove for now.

creativedrewy (commented):
Some replies

android/clientlib-ktx/src/main/java/com/solana/mobilewalletadapter/clientlib/MobileWalletAdapter.kt
```diff
@@ -0,0 +1,187 @@
...
+            withContext(ioDispatcher) {
+                client.authorize(identityUri, iconUri, identityName).get()
+                true
+            }
+        }
+    }
+
+    suspend fun reauthorize(identityUri: Uri, iconUri: Uri, identityName: String): Boolean {
+        return withScenario(associate()) { client ->
+            withContext(ioDispatcher) {
```
olivierg13, line 53: Something I notice just now, `withContext(ioDispatcher)` might be better at the root of this method (to execute `associate()` under the IO Dispatcher too)?

android/clientlib-ktx/src/main/java/com/solana/mobilewalletadapter/clientlib/MobileWalletAdapter.kt
```diff
@@ -0,0 +1,187 @@
...
+
+@Suppress("BlockingMethodInNonBlockingContext")
+class MobileWalletAdapter(
+    private val resultCaller: ActivityResultCaller,
+    private val scenarioType: ScenarioTypes = ScenarioTypes.Local,
+    private val timeout: Int = Scenario.DEFAULT_CLIENT_TIMEOUT_MS,
+    private val ioDispatcher: CoroutineDispatcher = Dispatchers.IO
+) {
+    private val contract = LocalAssociationIntentContract()
+    private val resultLauncher = resultCaller.registerForActivityResult(contract) { }
```
olivierg13, line 38: If I remember correctly, `registerForActivityResult()` needs to be called when the activity is being created (or at least before it's actually started/resumed, something like that).

Is there a risk that developers would instantiate `MobileWalletAdapter` later than this, leading to a potential exception?

android/clientlib-ktx/src/main/java/com/solana/mobilewalletadapter/clientlib/TransactionParams.kt
```diff
@@ -0,0 +1,24 @@
+package com.solana.mobilewalletadapter.clientlib
+
+import com.solana.mobilewalletadapter.common.ProtocolContract
+import com.solana.mobilewalletadapter.common.protocol.CommitmentLevel
+
+sealed class RpcCluster(val name: String) {
```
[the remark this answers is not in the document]
creativedrewy: This is a very good idea; we just want to enable custom versions as well. So maybe a janky ol' Java "Bean" class.

android/clientlib-ktx/src/main/java/com/solana/mobilewalletadapter/clientlib/MobileWalletAdapter.kt
creativedrewy: Let me read the docs on that.

creativedrewy: Welp...

android/clientlib-ktx/src/main/java/com/solana/mobilewalletadapter/clientlib/MobileWalletAdapter.kt
```diff
@@ -0,0 +1,119 @@
...
+    private val timeout: Int = Scenario.DEFAULT_CLIENT_TIMEOUT_MS,
+    private val ioDispatcher: CoroutineDispatcher = Dispatchers.IO
+) {
+    private val contract = LocalAssociationIntentContract(LocalAssociationIntentCreator())
+    private val resultLauncher = resultCaller.registerForActivityResult(contract) { }
+
+    private val scenarioChooser = ScenarioChooser()
+
+    suspend fun authorize(identityUri: Uri, iconUri: Uri, identityName: String): Boolean {
+        return withScenario(associate()) { client ->
```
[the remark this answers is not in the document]
olivierg13: Definitely, I can see the RxJava APIs using some of these pieces as well, and thus it could make sense to port at least some of these changes to the "core" clientlib.

PR #102, opened 2022-07-12
Adding Rx convenience MobileWalletAdapter APIs

olivierg13:
To match the coroutines APIs which are allowing devs to map the whole flow (start/action/close) into a single...Single 😁

### DIFF
--- a/android/clientlib-rxjava/build.gradle
+++ b/android/clientlib-rxjava/build.gradle
@@ -19,9 +19,11 @@ android {
 }
 
 dependencies {
-    compileOnly 'androidx.annotation:annotation:1.4.0'
+    compileOnly "androidx.annotation:annotation:1.4.0"
 
     api(project(":clientlib"))
 
     implementation "io.reactivex.rxjava2:rxjava:2.2.21"
+    implementation "io.reactivex.rxjava2:rxandroid:2.1.1"
+
 }
--- a/android/clientlib-rxjava/src/main/java/com/solana/mobilewalletadapter/clientlib/RxMobileWalletAdapter.java
+++ b/android/clientlib-rxjava/src/main/java/com/solana/mobilewalletadapter/clientlib/RxMobileWalletAdapter.java
@@ -1,0 +1,6 @@
+package com.solana.mobilewalletadapter.clientlib;
+
+public class RxMobileWalletAdapter {
+
+
+}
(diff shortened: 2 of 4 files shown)


android/clientlib-rxjava/src/main/java/com/solana/mobilewalletadapter/clientlib/RxMobileWalletAdapter.java
```diff
@@ -0,0 +1,158 @@
...
+        this.mRxLocalAssociationScenario = new RxLocalAssociationScenario(clientTimeoutMs);
+    }
+
+    @CheckResult
+    @NonNull
+    public Single<AuthorizeResult> authorize(@Nullable Uri identityUri,
+                                             @Nullable Uri iconUri,
+                                             @Nullable String identityName) {
+        return mRxLocalAssociationScenario
+                .start()
```
creativedrewy, line 38: Be still my beating heart!

creativedrewy (approved):
This is 🔥 !!!

android/clientlib-rxjava/build.gradle
```diff
@@ -6,7 +6,7 @@ android {
     compileSdk 32
 
     defaultConfig {
-        minSdk 23
+        minSdk 24
```
sdlaver, line 9: What necessitated this? It's not the end of the world, but it would be convenient to keep all the minSdk versions of MWA in lockstep.

android/clientlib-rxjava/src/main/java/com/solana/mobilewalletadapter/clientlib/RxMobileWalletAdapter.java
```diff
@@ -0,0 +1,117 @@
...
+    public RxMobileWalletAdapter(@IntRange(from = 0) int clientTimeoutMs) {
+        this.mRxLocalAssociationScenario = new RxLocalAssociationScenario(clientTimeoutMs);
+    }
+
+    @CheckResult
+    @NonNull
+    public Single<AuthorizeResult> authorize(@Nullable Uri identityUri,
+                                             @Nullable Uri iconUri,
+                                             @Nullable String identityName) {
+        return startExecuteAndClose(
```
sdlaver, line 40: Same comment as left on the Kotlin PR: https://github.com/solana-mobile/mobile-wallet-adapter/pull/101#discussion_r919466107

android/clientlib-rxjava/build.gradle
creativedrewy: Here is my thought on this matter. We are so close to API 24, I think we should just make the jump. That's still Android 7.1 which is very, very old. Saga is 6 OS versions ahead of that. If the community _needs_ lower API support, they can fork and make it work for themselves. Thoughts?

android/clientlib-rxjava/src/main/java/com/solana/mobilewalletadapter/clientlib/RxMobileWalletAdapter.java
creativedrewy: Yeah let's have a chat about it!

PR #107, opened 2022-07-13
fix: rename JS wallet API to more closely match upcoming Kotlin API

steveluscher:
We're moving toward the API being:

```
transact(walletAPI => { /* ... */ });
```

…to leave open the future for an API like

```
transact.usingRemoteWallet(remoteConfig, walletAPI => { /* ... */ });
````


creativedrewy (approved):
Transact the planet!!!

PR #111, opened 2022-07-15
Update MWA spec to 0.1.1 with wallet adapter feedback

sdlaver:


spec/spec.md
```diff
@@ -264,12 +265,15 @@ After [session establishment](#session-establishment) completes, the wallet endp
...
+- the AES-128-GCM message ciphertext, encrypted using the private key created during [session establishment](#session-establishment) and the random IV, and including the 4-byte message sequence number as additional authorized data (AAD)
 - the 16-byte authentication tag generated during AES-128-GCM encryption
 
 After decrypting the ciphertext with the shared secret generated during [session establishment](#session-establishment) and verifying the authentication tag, the receiving endpoint should further interpret it as a [JSON-RPC 2.0](https://www.jsonrpc.org/specification) message.
 
+The message sequence number is a monotonically increasing unsigned integer, which starts at 1 when session establishment completes. Each endpoint maintains its own independent sequence number, and increments it each time an encrypted message is created and sent. On receipt of an encrypted message, each endpoint should ensure that the sequence number is greater than that of the previous message received (other than for the first message received). If the value is less than or equal to the previously observed sequence number, the encrypted message should be discarded and the connection closed.
```
sdlaver, line 275: This provides resistance to replay attacks

spec/spec.md
```diff
@@ -436,6 +449,8 @@ where:
 
 This method attempts to renew the specified `auth_token`. The meaning of renew is an implementation detail of the wallet endpoint; it may return the same token, issue a new token, or refuse to renew the token. If the result is `ERROR_AUTHORIZATION_FAILED`, the token could not be renewed. A new token will need to be requested with the [`authorize`](#authorize) method. If a new `auth_token` is returned, the dapp endpoint should assume that the old `auth_token` is no longer valid and discard it.
 
+In addition to the reauthorized auth token, the wallet endpoint may optionally return updated values for `public_key` and `wallet_uri_base`, overriding any previously provided values. This allows a wallet endpoint to modify the set of authorized accounts for an already-issued auth token, without requiring the dapp endpoint to restart the authorization process.
+
```
sdlaver, line 453: Wallets have a few options here:
1. The new parameters to reauthorize can be used if the wallet wishes to automatically add or remove account access for previously issued auth tokens. Internally, it would set a flag on all modified auth tokens indicating that they require reauthorization. The next time they are attempted to be used, the dapp would get an `ERROR_REAUTHORIZE` result, would perform reauthorization, and receive updated accounts.
2. If a wallet wishes to force dapps to go through a full authorization process again, they can just invalidate all auth tokens in their internal store, and dapps will get an `ERROR_AUTHORIZATION_FAILED` result. At this point, the dapp should discard the cached auth token and re-run authorization
3. Some hybrid of the two - use `ERROR_REAUTHORIZE` to remove accounts, and `ERROR_AUTHORIZATION_FAILED` to add accounts (by requiring a new authorization.

spec/spec.md
```diff
@@ -760,10 +799,9 @@ const ERROR_INVALID_PAYLOAD = -3
 const ERROR_NOT_SIGNED = -4
 const ERROR_NOT_COMMITTED = -5
 const ERROR_TOO_MANY_PAYLOADS = -6
+const ERROR_CLUSTER_NOT_SUPPORTED = -7
 
 const ERROR_ATTEST_ORIGIN_ANDROID = -100
-
-const MAX_TRANSACTION_SZ = 1232
```
sdlaver, line 766: The MWA protocol doesn't need to define this explicitly. A wallet will refuse to sign an oversized transaction payload. Not having this in the MWA spec means one less magic constant to keep in sync.

creativedrewy (approved):

spec/spec.md
```diff
@@ -264,12 +265,15 @@ After [session establishment](#session-establishment) completes, the wallet endp
 
 After the [session establishment](#session-establishment) process completes, every message received by an endpoint is expected to be encrypted with AES-128-GCM (as specified by [NIST SP 800-38D](https://nvlpubs.nist.gov/nistpubs/Legacy/SP/nistspecialpublication800-38d.pdf)). The sending endpoint should prepare the encrypted message by concatenating:
 
+- the message sequence number, a 4-byte big-endian unsigned integer
 - a random 12-byte IV (which should be newly generated for each encrypted message)
-- the AES-128-GCM message ciphertext, encrypted using the private key created during [session establishment](#session-establishment) and the random IV
+- the AES-128-GCM message ciphertext, encrypted using the private key created during [session establishment](#session-establishment) and the random IV, and including the 4-byte message sequence number as additional authorized data (AAD)
```
steveluscher, line 270: Presumably I can supply this AAD as part of this SubtleCrypto API for the JS implementation?

```
interface AesGcmParams extends Algorithm {
    additionalData?: BufferSource; // <----
    iv: BufferSource;
    tagLength?: number;
}
```

spec/spec.md
```diff
@@ -310,29 +315,31 @@ where:
 - `dapp_uri`: (Optional) a URI representing the web address associated with the dapp endpoint making this authorization request. If present, it must be an absolute, hierarchical URI.
 - `dapp_icon_relative_path`: (Optional) a relative path (from dapp_uri) to an image asset file of an icon identifying the dapp endpoint making this authorization request
 - `dapp_name`: (Optional) the display name for this dapp endpoint
+- `cluster`: (optional) if set, the Solana network cluster with which the dapp endpoint intends to interact; supported values include `mainnet_beta`, `testnet`, `devnet`. If not set, the wallet endpoint determines which cluster to send transactions to.
```
steveluscher, line 318:
```suggestion
- `cluster`: (optional) if set, the Solana network cluster with which the dapp endpoint intends to interact; supported values include `mainnet-beta`, `testnet`, `devnet`. If not set, the wallet endpoint determines which cluster to send transactions to.
```

To keep consistent with the cluster type here:

https://github.com/solana-labs/solana/blob/ef30f1729c91e46833247f37173655cfded7e02f/web3.js/src/util/cluster.ts#L14

spec/spec.md
```diff
@@ -343,6 +350,8 @@ The lifetime of the returned `auth_token` is not defined. If a privileged method
 
 Dapp endpoints should make every effort possible to [verify the authenticity](#dapp-identity-verification) of the presented identity. While the `dapp_uri` parameter is optional, it is strongly recommended - without it, the wallet endpoint may not be able to verify the authenticity of the dapp.
 
+The `cluster` parameter allows the dapp endpoint to select a specific Solana cluster with which to interact. This is relevant for both [`sign_transaction`](#sign_transaction), where a wallet may refuse to sign transactions without a currently valid blockhash, and for [`sign_and_send_transaction`](#sign_and_send_transaction), where the wallet endpoint must know which cluster to submit the transactions to. This parameter would normally be used to select a cluster other than `mainnet_beta` for dapp development and testing purposes. Under normal circumstances, this field should be omitted, in which case the wallet endpoint is free to either automatically select an appropriate cluster, or to present the user with a cluster selection option.
```
steveluscher, line 353:
```suggestion
The `cluster` parameter allows the dapp endpoint to select a specific Solana cluster with which to interact. This is relevant for both [`sign_transaction`](#sign_transaction), where a wallet may refuse to sign transactions without a currently valid blockhash, and for [`sign_and_send_transaction`](#sign_and_send_transaction), where the wallet endpoint must know which cluster to submit the transactions to. This parameter would normally be used to select a cluster other than `mainnet-beta` for dapp development and testing purposes. Under normal circumstances, this field should be omitted, in which case the wallet endpoint is free to either automatically select an appropriate cluster, or to present the user with a cluster selection option.
```

spec/spec.md
```diff
@@ -419,12 +428,16 @@ where:
 ```
 {
     “auth_token”: “<auth_token>”,
+    “pub_key”: [“<public_key>", ...]
```
steveluscher, line 431: Should we pluralize this now?
```suggestion
    “pub_keys”: [“<public_key>", ...]
```

spec/spec.md
```diff
@@ -310,29 +315,31 @@ where:
...
+- `cluster`: (optional) if set, the Solana network cluster with which the dapp endpoint intends to interact; supported values include `mainnet_beta`, `testnet`, `devnet`. If not set, the wallet endpoint determines which cluster to send transactions to.
 
 ###### Result
 {: .no_toc }
 
 ```
 {
     “auth_token”: “<auth_token>”,
-    “pub_key”: “<public_key>
+    “pub_key”: [“<public_key>", ...]
```
steveluscher, line 326:
```suggestion
    “pub_keys”: [“<public_key>", ...]
```

spec/spec.md
```diff
@@ -419,12 +428,16 @@ where:
...
     “auth_token”: “<auth_token>”,
+    “pub_key”: [“<public_key>", ...]
+    “wallet_uri_base”: “<wallet_uri_base>”,
 }
 ```
 
 where:
 
 - `auth_token`: as defined for [`authorize`](#authorize)
+- `public_key`: (optional) if set, as defined for [`authorize`](#authorize). If not set, the previously provided `public_key` values should be retained.
```
steveluscher, line 439:
```suggestion
- `pub_keys`: (optional) if set, as defined for [`authorize`](#authorize). If not set, the previously provided `pub_keys` values should be retained.
```

spec/spec.md
```diff
@@ -436,6 +449,8 @@ where:
 
 This method attempts to renew the specified `auth_token`. The meaning of renew is an implementation detail of the wallet endpoint; it may return the same token, issue a new token, or refuse to renew the token. If the result is `ERROR_AUTHORIZATION_FAILED`, the token could not be renewed. A new token will need to be requested with the [`authorize`](#authorize) method. If a new `auth_token` is returned, the dapp endpoint should assume that the old `auth_token` is no longer valid and discard it.
 
+In addition to the reauthorized auth token, the wallet endpoint may optionally return updated values for `public_key` and `wallet_uri_base`, overriding any previously provided values. This allows a wallet endpoint to modify the set of authorized accounts for an already-issued auth token, without requiring the dapp endpoint to restart the authorization process.
```
steveluscher, line 452:
```suggestion
In addition to the reauthorized auth token, the wallet endpoint may optionally return updated values for `pub_keys` and `wallet_uri_base`, overriding any previously provided values. This allows a wallet endpoint to modify the set of authorized accounts for an already-issued auth token, without requiring the dapp endpoint to restart the authorization process.
```

spec/spec.md
```diff
@@ -703,13 +740,15 @@ sign_message
...
     “auth_token”: “<auth_token>”,
+    "pub_key": "<public_key>",
     “payloads”: [“<message>”, ...],
 }
 ```
 
 where:
 
 - `auth_token`: an auth_token returned by [`authorize`](#authorize), [`reauthorize`](#reauthorize), or [`clone_authorization`](#clone_authorization) for which access to `sign_message` was requested
+- `public_key`: (optional) if specified, the base64-encoded public key of the account which should be used to sign `message`. This should be one of the public keys returned by `authorize` for `auth_token`. If not specified, the wallet endpoint will determine which account to use to sign `message`.
```
steveluscher, line 751:
```suggestion
- `pub_key`: (optional) if specified, the base64-encoded public key of the account which should be used to sign `message`. This should be one of the public keys returned by `authorize` for `auth_token`. If not specified, the wallet endpoint will determine which account to use to sign `message`.
```

steveluscher (approved):
<3

spec/spec.md
jordansexton: > Under normal circumstances, this field should be omitted, in which case the wallet endpoint is free to either automatically select an appropriate cluster, or to present the user with a cluster selection option.

This being optional is weird because it reduces the explicitness of the API. Why would the dapp want to leave the selection up to the wallet in any circumstance? It knows where it wants to send transactions, and it needs to confirm them against an RPC node on the same cluster.

It also doesn't get a value back from the wallet to indicate which cluster the wallet chose, leaving it to guess or try to infer from the recent blockhash?

jordansexton:
Using `address` instead of `pub_key` would allow this to work for other chains which primarily use truncated/check summed hashes of pubkeys for accounts rather than pubkeys directly.

spec/spec.md
```diff
@@ -607,19 +644,19 @@ sign_and_send_transaction
...
     “payloads”: [“<transaction>”, ...],
-    “commitment”: “<commitment_level>”,
-    "cluster": "<cluster>",
-    "skip_preflight": <skip_preflight>,
-    "preflight_commitment": "<preflight_commitment>",
+    "options": {
+        “commitment”: “<commitment_level>”,
+        "skip_preflight": <skip_preflight>,
+        "preflight_commitment": "<preflight_commitment>",
+    }
```
jordansexton, line 651: How important are the `options` (`commitment_level`, `skip_preflight`, `preflight_commitment`)? Are wallets required to honor these? What are the implications of this? For example, do wallets actually want you daps be able to control the preflight (simulation) commitment, or skipping [preflight transaction checks](https://docs.solana.com/developing/clients/jsonrpc-api#sendtransaction) (what exactly does this do?)

jordansexton:
This is a little bit bikesheddy but might as well put it to bed --

Why do we like `sign_transaction` / `sign_and_send_transaction` / `sign_message` with multiple inputs / outputs instead of plural names for these? It's also a bit confusing if you get the return value and it's `signed_transaction` / `signature` but it's actually multiple items.

The way I'd probably approach this is as a caller is to destructure and iterate over the it (e.g. `for (signed_transaction of signed_transactions)`) which this naming ends up being a bit awkward for.

I can see an argument for the method names being singular (fewer keystrokes, plus it doesn't matter) but the returns being plural (`signature` representing an array of signatures is pretty weird).

spec/spec.md
sdlaver: Yep! There's a draft PR up for this with partial JS implementation here: https://github.com/solana-mobile/mobile-wallet-adapter/pull/112

spec/spec.md
sdlaver: None of the other terms (`transaction`, `message`) are pluralized in method call parameters. It's primarily intended for machine consumption, so I don't know that it matters all that much, but happy to change them all for consistency.

sdlaver:
> ... plural names ...

It just evolved that way - all these were originally single values, and then became plurals with iteration on the spec. I'll change them all for logical consistency.

spec/spec.md
steveluscher: I commented on #112, but: why less than or equal to? Why not ‘exactly the next number in sequence?’

sdlaver: Left my comments on this topic here: https://github.com/solana-mobile/mobile-wallet-adapter/pull/112#discussion_r922849562

spec/spec.md
sdlaver: Good point. Making the default `mainnet-beta`

spec/spec.md
sdlaver: The preflight options were requested by a wallet developer, to align with the `sendTransaction` options of web3.js and the Solana RPC interface. `commitment` provides the dapp with some control over the finality at which the transaction would be considered successfully sent.

Re: skipping preflight checks, it's in the API description you linked:
> Before submitting, the following preflight checks are performed:
> 
> The transaction signatures are verified
> The transaction is simulated against the bank slot specified by the preflight commitment. On failure an error will be returned. Preflight checks may be disabled if desired. It is recommended to specify the same commitment and preflight commitment to avoid confusing behavior.

Re: including these options, I think I'd pose it as the inverse - our other methods to submit transactions to the RPC nodes (web3.js, RPC API, solana-client) have preflight control options. Is there a good reason to not have them here? I'd not want for a MWA implementation of a dapp to be less capable than a wallet-adapter + web3.js implementation of a dapp.

spec/spec.md
sdlaver: Done

spec/spec.md
sdlaver: Done

spec/spec.md
sdlaver: Replaced with `accounts` instead

spec/spec.md
sdlaver: Replaced with `accounts` instead

spec/spec.md
sdlaver: Replaced with `accounts` instead

spec/spec.md
sdlaver: Replaced with `accounts` instead

spec/spec.md
sdlaver: Replaced with `accounts` instead

sdlaver:
> Using `address` instead of `pub_key` would allow this to work for other chains which primarily use truncated/checksummed hashes of pubkeys for accounts rather than pubkeys directly.

I ended up using `account` instead. Any strong preference for `address`?

jordansexton:
> I ended up using `account` instead. Any strong preference for `address`?

Yes, I have a strong preference for `address`.

`address` is the widely used term on Bitcoin, Cosmos, and EVM chains which don't use public keys directly as account identifiers. JS wallet libraries targeting the EVM [like Ethers](https://docs.ethers.io/v5/api/signer/#Wallet--properties) expose both an `address` and `publicKey`, since addresses identify accounts but you may still need the public key the address is derived from.

Also, `account` on Solana [refers to an account](https://docs.solana.com/terminology#account) referenced by a pubkey (aka key or address). Calling keys, pubkeys, or addresses themselves accounts on Solana is a historical source of confusion.

sdlaver:
> Yes, I have a strong preference for `address`.

Updated to use `address` or `addresses`, as appropriate

spec/spec.md
jordansexton: > Is there a good reason to not have them here?

I can think of a few reasons (goodness debatable).

1. Keep initial versions of the API simple.

It's easier to add stuff later than to remove it. The use case for having the dapp provide these is unclear to me. We can always add these options if people really need them. Did the wallet partner provide any guidance about how they expect it to be used?

Also, the longer the wallet waits for confirmation, the longer the dapp will get backgrounded for. Does this matter?

2. The wallet should control the interaction and user security.

This is a bit opinionated, but hear me out.

When a dapp has an RPC connection of its own, it makes sense for it to configure these values. When it's asking the wallet to send things, it makes sense for the wallet to configure these values. Once the wallet gets the signature, from the RPC and returns control to the dapp, the dapp can always wait for a status of `finalized` if it wants to.

Preflight control from the dapp is also not as important if the wallet is simulating the transaction as a distinct step (which it should do!) before having the user sign and submit it. What exactly is the dapp asking to do when it overrides the wallet's defaults? Should wallets be encouraged to let this happen?

3. Anticipate a chain agnostic design.

It seems like multi chain support is going to be important (people are already asking about it). These options are the only very Solana-specific thing in the API. Arguably the network/cluster name is, but the cluster names could be interpreted to mean `solana:<cluster>`.

Building off point 1, it seems better to add this kind of thing later once we have the future API more clearly in mind, unless we know we need it.

jordansexton: One other point --

> I'd not want for a MWA implementation of a dapp to be less capable than a wallet-adapter + web3.js implementation of a dapp.

While it's true that WalletAdapter _currently_ has these `SendOptions` and forwards them to browser wallets that accept them, they have been removed from the wallet standard draft for the same reasons as above.

Once a dapp has the broadcast transaction signature from the wallet, it can wait longer than the wallet generally wants to, but not less long. This seems right to me -- the wallet owns the UX when sending, so it should own the configuration.

spec/spec.md
jordansexton: > Under normal circumstances, this field should be omitted, in which case the wallet endpoint is free to either automatically select an appropriate cluster, or to present the user with a cluster selection option.

How does the dapp find out which cluster the wallet chose?

sdlaver: This comment is on an old version of the changes. The latest text reads:
> This parameter would normally be used to select a cluster other than `mainnet-beta` for dapp development and testing purposes. Under normal circumstances, this field should be omitted, in which case the wallet endpoint will interact with the `mainnet-beta` cluster.

jordansexton: Ahh, my bad. Thanks!

spec/spec.md
```diff
@@ -302,46 +306,53 @@ authorize
...
 where:
 
-- `dapp_uri`: (Optional) a URI representing the web address associated with the dapp endpoint making this authorization request. If present, it must be an absolute, hierarchical URI.
-- `dapp_icon_relative_path`: (Optional) a relative path (from dapp_uri) to an image asset file of an icon identifying the dapp endpoint making this authorization request
-- `dapp_name`: (Optional) the display name for this dapp endpoint
+- `identity`: a JSON object, containing:
+  - `uri`: (optional) a URI representing the web address associated with the dapp endpoint making this authorization request. If present, it must be an absolute, hierarchical URI.
+  - `icon`: (optional) a relative path (from `uri`) to an image asset file of an icon identifying the dapp endpoint making this authorization request
+  - `name`: (optional) the display name for this dapp endpoint
+- `cluster`: (optional) if set, the Solana network cluster with which the dapp endpoint intends to interact; supported values include `mainnet-beta`, `testnet`, `devnet`. If not set, defaults to `mainnet-beta`.
```
jordansexton, line 319: `cluster` is Solana-specific, would we consider `network` here? If the API later comes to allow a [chain ID](https://github.com/ChainAgnostic/CAIPs/blob/master/CAIPs/caip-2.md) of some kind, `network` could still be multi-chain.

spec/spec.md
sdlaver: > Did the wallet partner provide any guidance about how they expect it to be used?

I will ask for more context on why they want this

> Preflight control from the dapp is also not as important if the wallet is simulating the transaction as a distinct step

I agree, the case for independent preflight control is shaky. Happy to revisit this with the wallet.

> Also, the longer the wallet waits for confirmation, the longer the dapp will get backgrounded for. Does this matter?

I'd argue that the right thing to do here is to keep the dapp backgrounded until the transaction reaches a desired level of confirmation (in either case, whether the wallet decides the appropriate level or the dapp does). This way, the user gets a consistent "sign and send" experience - they always enter the wallet, approve the transaction, and wait in a wallet UI context until the transaction reaches the desired confirmation, after which they return to the dapp. I say that this UX is desirable because it doesn't leave the "doing the right thing" part to the dapps.

_If_ we still want to allow dapps to have a say in this behavior, keeping just the `commitment` field allows them to configure what level of finality they are comfortable with, and implement their own in-dapp waiting experience by selecting `processed`.

sdlaver: Also,

> These options are the only very Solana-specific thing in the API

This is covered in the new "Support for other chains" section at the end. The shape of the `options` object would differ from chain to chain. You could imagine that, in a future rev of the spec that fully specified a second chain, the shape of this object would be extracted and put into a Solana-specific companion spec. The name of the field, `options`, would remain the same, with its shape controlled by the type of association (which encodes the chain in use, e.g. `solana-wallet://`)

spec/spec.md
sdlaver: Same answer as my previous comment re: `options`.

IMO, every chain is going to have a few unique things, which are best captured in chain-specific companion specs. For now, since v1.0 is a Solana-only spec, they are encoded right into the main spec, but I envision this cluster field as being one of those.

spec/spec.md
sdlaver: I discussed with the wallet, and they are OK with dropping the `preflight` parameters from `sign_and_send_transaction`. This will leave only `commitment` in `options`. PR has been updated.

spec/spec.md
```diff
@@ -302,46 +306,53 @@ authorize
...
+Dapp endpoints should make every effort possible to [verify the authenticity](#dapp-identity-verification) of the presented identity. While the `uri` parameter is optional, it is strongly recommended - without it, the wallet endpoint may not be able to verify the authenticity of the dapp.
+
+The `cluster` parameter allows the dapp endpoint to select a specific Solana cluster with which to interact. This is relevant for both [`sign_transactions`](#sign_transactions), where a wallet may refuse to sign transactions without a currently valid blockhash, and for [`sign_and_send_transactions`](#sign_and_send_transactions), where the wallet endpoint must know which cluster to submit the transactions to. This parameter would normally be used to select a cluster other than `mainnet-beta` for dapp development and testing purposes. Under normal circumstances, this field should be omitted, in which case the wallet endpoint will interact with the `mainnet-beta` cluster.
```
steveluscher, line 355: I wonder; do we actually expect wallets to check the validity of a blockhash before signing, or is that not typical. I guess the words ‘may refuse’ covers this. Maybe make that even looser via ‘may choose to refuse?’

Or maybe take it out altogether; if I pass a wallet a transaction that includes a _durable nonce_ value, there's nothing that they can do to verify it's still valid unless I _also_ pass them the address of the nonce account.

Just trying to prevent the spec from writing cheques it can't cash here.

steveluscher: Wallets should probably just shut up and sign transactions. Maybe we should take `cluster` out here.

steveluscher: cc/ @jordansexton.

spec/spec.md
```diff
@@ -607,21 +646,18 @@ sign_and_send_transaction
...
-- `cluster`: (optional) if set, the Solana network cluster to send the signed transaction(s) to; supported values include `mainnet_beta`, `testnet`, `devnet`. If not set, the wallet endpoint determines which cluster to send transactions to.
-- `skip_preflight`: (optional) see the JSON-RPC [sendTransaction](https://docs.solana.com/developing/clients/jsonrpc-api#sendtransaction) API. If not set, defaults to `false`.
-- `preflight_commitment`: (optional) see the JSON-RPC [sendTransaction](https://docs.solana.com/developing/clients/jsonrpc-api#sendtransaction) API. If not set, defaults to `finalized`.
+- `auth_token`: an auth_token returned by [`authorize`](#authorize), [`reauthorize`](#reauthorize), or [`clone_authorization`](#clone_authorization) for which access to `sign_and_send_transactions` was requested
+- `payloads`: one or more base64-encoded transaction payload to sign
+- `options`: a JSON object, containing:
+  - `commitment`: (optional) if set, one of `processed`, `confirmed`, or `finalized`, specifying the desired commitment level that must be reached for this request to be successful. If not set, defaults to `confirmed`.
```
steveluscher, line 660: So, hold on, I've been meaning to ask this. Are we actually making the wallets do transaction confirmation? This would be a new thing, for wallet adapters.

Take, for instance, the current Phantom wallet adapter for desktop web:

https://github.com/solana-labs/wallet-adapter/blob/master/packages/wallets/phantom/src/adapter.ts#L209-L210

It sends the transaction, returns the (unasserted) signature, and then leaves confirmation up to the dApp.

* Control returns to the dApp right away
* Users can perform new orthogonal transactions while waiting for the last one(s) to complete
* The dApp can care about the confirmation or not (eg. maybe it's no longer important if the user has navigated away from that screen)

steveluscher (changes requested):
Awesome! I was going to accept-with-nits this, but maybe we have to have a conversation about in-wallet transaction confirmation, and whether we want to be in that game or not.

spec/spec.md
sdlaver: > Wallets should probably just shut up and sign transactions

The wallet needs to know this, so it knows where to send a transaction. I think the only debate is whether this should be a property of `sign_and_send_transaction`, or of `authorize`. Putting it here allows the wallet to make better decisions (but does not force it to do so).

spec/spec.md
sdlaver: I haven't had push-back from wallets on this, but it's possible that the implications of this haven't settled in yet. Let's try and force the issue, and see what wallets say.

From a UX standpoint, I can see benefits to making the transaction confirmation experience common across dapps. However, it does differ from the other implementations of this functionality.

It's worth noting that this gives the _option_ to the dapp of having the in-wallet confirmation experience, but does not require it. They can specify `processed` here and get control back very quickly, and then perform their own wait UX, or continue on with their day, etc.

spec/spec.md
steveluscher: Oh no, I wasn't talking about `sign_and_send_transaction`. I was talking about `sign_transaction`. You can definitely sign without knowing the cluster.

steveluscher: It's this bit that I'm stuck on.

> This is relevant for both [`sign_transactions`](#sign_transactions), where a wallet may refuse to sign transactions without a currently valid blockhash

sdlaver: That was a suggestion I got from @jordansexton. Let's wait for him to weigh in.

spec/spec.md
sdlaver: We could always add some indication to not wait at all (a `none` value, perhaps?), but I have some wallet feedback requests out at the moment, and would like that input before we invest additional time in the details here.

steveluscher: Found it. From the docs for the [`sendTransaction` RPC method](https://docs.solana.com/developing/clients/jsonrpc-api#sendtransaction):

> If the node's rpc service receives the transaction, this method immediately succeeds, without waiting for any confirmations. A successful response from this method does not guarantee the transaction is processed or confirmed by the cluster.

sdlaver: Given that this PR isn't adding the `commitment` parameter, I'm going to propose that we break this sub-issue off from this PR. That'll unblock work on the agreed parts of this spec, and we can tackle the discussions with wallets on this point separately.

jozanza: Hey Phantom dev here! 👋🏽 

@sdlaver I think that's a good call. `sign_and_send_transactions` has some pretty large implications when it comes to tx simulation inside of the wallet approval notification. There is also the concept of JSON RPC batching to consider.

Generally, I'd also like to understand where this spec aligns and deviates from the standard wallet adapter interface. In an ideal world, this RPC spec is the same for any portions that use the same method. I hear there's a Discord server opening soon where I can maybe get a bit more looped in. In the meantime, I'd appreciate any links to important conversations deviations in this RPC spec vs the existing one. Thanks, y'all 🙏🏽 ❤️

sdlaver: Hi @jozanza, we have https://discord.gg/solanamobile for all discussions SMS-related. There's a #wallet channel there (which reminds me, I still need to point to this PR there). Let's have a discussion on where the spec aligns, and where it differs, in that forum.

jozanza: Sounds like a plan. I'll see you there. Thanks!

sdlaver: Filed https://github.com/solana-mobile/mobile-wallet-adapter/issues/134 to track closing on this discussion separately from this PR.

steveluscher (approved):

PR #112, opened 2022-07-15
Implement message sequence numbers for native + JS implementations

sdlaver:


js/packages/mobile-wallet-adapter-protocol/src/jsonRpcMessage.ts
```diff
@@ -11,24 +12,29 @@ type JSONRPCResponse<TMessage> = {
 
 export async function encryptJsonRpcMessage(jsonRpcMessage: unknown, sharedSecret: SharedSecret) {
     const plaintext = JSON.stringify(jsonRpcMessage);
+    const sequenceNumber = new Uint8Array(SEQUENCE_NUMBER_BYTES);
     const initializationVector = new Uint8Array(INITIALIZATION_VECTOR_BYTES);
+    // TODO: populate sequence number (big-endian)
```
sdlaver, line 17: @steveluscher could use your advice or assistance on these TODOs. I'm introducing the concept of message sequence numbers (see related MWA spec PR for details). With my weak JS skills, I'm not sure where the best place to track these values is, nor the best way to inject them into this function (parameters? globals?).

android/common/src/main/java/com/solana/mobilewalletadapter/common/protocol/MobileWalletAdapterSessionCommon.java
```diff
@@ -39,6 +40,7 @@
 public abstract class MobileWalletAdapterSessionCommon implements MessageReceiver, MessageSender {
     private static final String TAG = MobileWalletAdapterSessionCommon.class.getSimpleName();
 
+    private static final int SEQ_NUM_LENGTH_BYTES = 4;
```
steveluscher, line 43: A ceiling of 16 commands per session _feels_ too low to me, no?

steveluscher: Oh bytes. LOL. Nevermind.

android/common/src/main/java/com/solana/mobilewalletadapter/common/protocol/MobileWalletAdapterSessionCommon.java
```diff
@@ -195,13 +205,20 @@ protected byte[] decryptSessionPayload(@NonNull byte[] payload) throws SessionMe
             throw new IllegalStateException("Cannot decrypt, no session key has been established");
         }
 
+        final int seqNum = ByteBuffer.wrap(payload, 0, SEQ_NUM_LENGTH_BYTES).getInt(); // Big-endian
+        if (seqNum <= mSeqNumberRx) {
```
steveluscher, line 209: This would allow skipping a sequence number. Do we want to allow that?

steveluscher: If yes, cool! If not, then the variable we should track is more like `mNextSeqNumberRx` and then do…

```
if (seqNum != mNextSeqNumberRx) {
  throw...
}
mNextSeqNumberRx++;
```

sdlaver: I spent some time thinking about whether to shoot for strict sequence numbers, or permissive sequence numbers. I ended up deciding on permissive sequence numbers, because the spec is designed to support multiple different transports, and it doesn't set any hard requirements on whether or not a transport must be lossless (a la Websockets over TCP). Monotonically increasing sequence numbers hardens against replay attacks, but also allows for transports that:
- may fail in a way that the sender can observe. While we could roll back the sequence number in this case, it adds complexity and inter-module coupling to the sender's encryption implementation
- may fail in a way that the sender cannot observe. In this case, while one operation may fail, the channel remains resilient. I debated whether this is desirable behavior, since a failure to send implies a failure to provide service, but ultimately decided that this is already a property of a lossy transport, and adding sequence numbers can't fix this, so sequence numbers should be tolerant of it (for any application of the protocol that chooses to use a lossy transport method).

android/common/src/main/java/com/solana/mobilewalletadapter/common/protocol/MobileWalletAdapterSessionCommon.java
```diff
@@ -195,13 +205,20 @@ protected byte[] decryptSessionPayload(@NonNull byte[] payload) throws SessionMe
...
+            throw new SessionMessageException("Encrypted messages has invalid sequence number");
```
steveluscher, line 210: Grammar nit.

steveluscher:
Added the JS implementation to your branch, @sdlaver. `git pull` it.

js/packages/mobile-wallet-adapter-protocol/src/transact.ts
```diff
@@ -51,6 +57,7 @@ export async function transact<TReturn>(
     const sessionPort = await startSession(associationKeypair.publicKey, config?.baseUri);
     const websocketURL = `ws://localhost:${sessionPort}/solana-wallet`;
     let nextJsonRpcMessageId = 1;
```
steveluscher, line 59: I just decided to hijack the JSON RPC message id as the sequence number.

steveluscher (approved):

js/packages/mobile-wallet-adapter-protocol/src/jsonRpcMessage.ts
```diff
@@ -1,34 +1,53 @@
...
-        getAlgorithmParams(initializationVector),
+        getAlgorithmParams(sequenceNumberVector, initializationVector),
         sharedSecret,
         Buffer.from(plaintext),
     );
-    const response = new Uint8Array(initializationVector.byteLength + ciphertext.byteLength);
-    response.set(new Uint8Array(initializationVector), 0);
-    response.set(new Uint8Array(ciphertext), initializationVector.byteLength);
+    const response = new Uint8Array(
+        sequenceNumberVector.byteLength + initializationVector.byteLength + ciphertext.byteLength,
```
steveluscher, line 34: I tested this PR with the examples. That's how I found this bug :)

js/packages/mobile-wallet-adapter-protocol/src/jsonRpcMessage.ts
```diff
@@ -1,40 +1,53 @@
...
+        getAlgorithmParams(sequenceNumberVector, initializationVector),
         sharedSecret,
         Buffer.from(plaintext),
     );
-    const response = new Uint8Array(initializationVector.byteLength + ciphertext.byteLength);
-    response.set(new Uint8Array(sequenceNumber), 0);
-    response.set(new Uint8Array(initializationVector), sequenceNumber.byteLength);
-    response.set(new Uint8Array(ciphertext), sequenceNumber.byteLength + initializationVector.byteLength);
+    const response = new Uint8Array(
+        sequenceNumberVector.byteLength + initializationVector.byteLength + ciphertext.byteLength,
```
steveluscher, line 34: I tested this PR with the examples. That's how I found this bug :)

olivierg13 (approved):

sdlaver:
Re-tested after updates with both `fakedapp` and `example-web-app`

PR #118, opened 2022-07-17
Remove auth_token from privileged methods; make the session authorization stateful instead

sdlaver:


sdlaver:
@steveluscher @jordansexton @olivierg13 @creativedrewy FYI

This PR still needs:
- Spec updates
- Updates to all corresponding client bindings

Overall, I'm happy with it. There'll be a future PR to clean up the wallet auth token management APIs a bit, but that was planned irrespective of this.

android/fakedapp/src/main/java/com/solana/mobilewalletadapter/fakedapp/MainViewModel.kt
```diff
@@ -91,6 +91,10 @@ class MainViewModel(application: Application) : AndroidViewModel(application) {
         }
 
         val signedTransactions = localAssociateAndExecute(sender) { client ->
+            val authorized = doReauthorize(client)
```
creativedrewy, line 94: So if I'm grokking this right, the clients will need to essentially make an authorize call before each operation? The assumption being that in many cases, the auth state will still be valid and thus will come back as authorized=true very quickly?

creativedrewy: And on a related note, our convenience wrappers are going to largely abstract this away, no?

creativedrewy (commented):
I like it! Just want to make sure I'm understanding it :)

android/fakedapp/src/main/java/com/solana/mobilewalletadapter/fakedapp/MainViewModel.kt
olivierg13: +1 that looks good to me if we do a `reauthorize()` before each operation, it's easy to abstract away in the convenience lib as well.

I'm guessing we wouldn't need to expose `reauthorize()` at all in the convenience libs, just `authorize()`?

olivierg13 (approved):

android/fakedapp/src/main/java/com/solana/mobilewalletadapter/fakedapp/MainViewModel.kt
sdlaver: Yep, after a session is established, it'll need either an `authorize` or a `reauthorize` call before any of the privileged method calls. I expect the convenience wrappers can abstract this away from consumers

creativedrewy: Yeah this sounds great. I'll wait to what you do with my precious ktx lib before approving :)

sdlaver:
PR https://github.com/solana-mobile/mobile-wallet-adapter/pull/123 contains related spec changes for stateful authorization

creativedrewy:
SHIP IT

sdlaver:
> SHIP IT

I'm going to wait for #123 to be merged, which itself is waiting on #111. I should be able to get them all in this week.

steveluscher:
All rebased and ready to verify!

creativedrewy (approved):
GTG

android/common/src/main/java/com/solana/mobilewalletadapter/common/ProtocolContract.java
```diff
@@ -64,12 +59,12 @@ public class ProtocolContract {
 
     public static final String RESULT_SIGNED_PAYLOADS = "signed_payloads"; // type: JSON array of String (base64-encoded signed payloads)
 
-    public static final int ERROR_REAUTHORIZE = -1;
-    public static final int ERROR_AUTHORIZATION_FAILED = -2;
-    public static final int ERROR_INVALID_PAYLOADS = -3;
-    public static final int ERROR_NOT_SIGNED = -4;
-    public static final int ERROR_NOT_COMMITTED = -5;
-    public static final int ERROR_TOO_MANY_PAYLOADS = -6;
+    // Keep these in sync with `mobile-wallet-adapter-protocol/src/errors.ts`.
```
sdlaver, line 62: 🥇

PR #149, opened 2022-07-21
A demonstration of connect-free app architecture, in React Native.

steveluscher:
# Premise

One of the incredibly powerful things about the mobile wallet adapter architecture is that you can issue _multiple directives_ to a wallet in a _single_ session. One of these directives is to *authorize the app in the first place*.

This means that you can create a truly *connect-free* application. A user can not have a wallet connected _at all_, perform an action in the app that requires authorization, and the app can decide whether:

1. They have an existing authorization that can be reused, or
2. Whether to authorize the app in the same session in which the privileges method is called. 

# Summary of change

* Eliminated the ‘connect’ button.
* Clicking ‘record’ or ’sign’ now
  * Authorizes the dApp if it hasn't been authorized yet, or
  * Reuses the cached authorization if it has.

# Demo

Look ma, no connect button!

https://user-images.githubusercontent.com/13243/180129478-e2434d95-4fe6-49ee-ae6c-8bf8be4ba667.mp4

### DIFF
--- a/examples/example-react-native-app/components/AccountBalance.tsx
+++ b/examples/example-react-native-app/components/AccountBalance.tsx
@@ -2,7 +2,7 @@ import {useConnection} from '@solana/wallet-adapter-react';
 import {LAMPORTS_PER_SOL, PublicKey} from '@solana/web3.js';
 import React, {useCallback, useMemo} from 'react';
 import {StyleSheet, View} from 'react-native';
-import {Subheading, Text} from 'react-native-paper';
+import {Headline, Text} from 'react-native-paper';
 import useSWR from 'swr';
 
 type Props = Readonly<{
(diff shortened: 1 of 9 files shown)


creativedrewy:
So this is pretty stinkin' cool, but I will say that users are possibly going to expect to see thier wallet address that they connected with. So perhaps the app could cache the address and show it after authorizing? For first time uses, you _could_ have a button that says "click to authorize & get address" or whatever, but it's just an option.

I feel like this sort of nudges into a greater UX discussion around security/safety. I think there is some level of safety - at least it _feels_ safe - to have to manually choose to authorize a site, and only then can you do transactions. Auth'ing and signing in one fell swoop could lead to users button mashing and getting into trouble. But then, maybe I'm wrong and it really isn't any safer. Just trying to think through the ramifications.

steveluscher:
It does cache and show it after authorizing! Check out the video.

creativedrewy:
Hah hah, this is what I get for reviewing PRs at like 11:30 pm

dFohlen:
@steveluscher does this also work with other wallets such as Phantom? I'm currently getting the message "Found no installed wallet that supports the mobile wallet protocol".

steveluscher:
It will! Right now there are no Mobile Wallet Adapter compatible wallets except for `fakewallet`, which is in this repository, but a number of wallet manufacturers are working on adding support right now!

In  the meantime you can build and install `fakewallet` for testing.

PR #157, opened 2022-07-22
Replace usage of base58 in the protocol with base64

### LINKED ISSUE #142: Remove base58 from the protocol in lieu of base64
This will align with spec changes made in #111

sdlaver:
Fixes #142
base58 should be used only with:
a. the presentation layer, when displaying signatures and public keys to a user
b. Solana JSON RPC node APIs, some of which require base58-encoded values


sdlaver:
@steveluscher when ready, can you drop the corresponding JS changes in here?

PR #162, opened 2022-07-23
Increase total connection backoff schedule to 30s

### LINKED ISSUE #82: Please help me answer another question. Thank you very much.
In the Android demo, when DAPP starts to authorize the connection, after pulling up the wallet application, it delays sending messages to the wallet for 100 milliseconds, and then the connection timeout is 20 milliseconds. I think the 100 millisecond delay is too short, and the 20 millisecond connection timeout is too short. Because many wallets are flutter applications, the time to start an application is far more than 100 milliseconds, and the connection timeout is 20 milliseconds, which is not enough for some applications and scenarios. This is the problem I encountered when accessing your SDK.

sdlaver:
Fixes #82

The connection timeout needs to be substantially increased, for two reasons:
1. Some wallets take longer than the current timeout (3.1s with the current
   clientlib implementation) to start and being listening for websocket
   connections. Data reported by several wallets indicate that 5-6 seconds is
   a more realistic startup time number for a cold start of wallets built using
   frameworks like React Native or Dart+Flutter.
2. If multiple wallets are installed AND the user has not chosen a default, the
   time that the user spends in the disambiguation dialog selecting a wallet
   needs to be included in this timeout as well.

This fix both extends the timeout, and also adds logic to the fakedapp sample to
short-circuit this connection timeout to 5s after the invoked wallet app
terminates.

### DIFF
--- a/js/packages/mobile-wallet-adapter-protocol/src/transact.ts
+++ b/js/packages/mobile-wallet-adapter-protocol/src/transact.ts
@@ -13,7 +13,6 @@ import { startSession } from './startSession';
 import { AssociationKeypair, MobileWallet, WalletAssociationConfig } from './types';
 
 const WEBSOCKET_CONNECTION_CONFIG = {
-    maxAttempts: 34,
     /**
      * 300 milliseconds is a generally accepted threshold for what someone
      * would consider an acceptable response time for a user interface
@@ -24,7 +23,8 @@ const WEBSOCKET_CONNECTION_CONFIG = {
      * user to be presented with a disambiguation dialog, select a wallet, and
      * for the wallet app to subsequently start.
      */
-    retryDelayMs: [150, 150, 200, 500, 500, 750, 750, 1000],
+    retryDelayScheduleMs: [150, 150, 200, 500, 500, 750, 750, 1000],
+    timeoutMs: 30000,
 } as const;
 const WEBSOCKET_PROTOCOL = 'com.solana.mobilewalletadapter.v1';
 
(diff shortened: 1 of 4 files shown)


js/packages/mobile-wallet-adapter-protocol/src/transact.ts
```diff
@@ -17,11 +17,14 @@ const WEBSOCKET_CONNECTION_CONFIG = {
...
+     * after having performed a low-attention tapping task. We set the initial
      * interval at which we wait for the wallet to set up the websocket at
-     * half this, as per the Nyquist frequency.
+     * half this, as per the Nyquist frequency, with a progressive backoff
+     * sequence from there. The total wait time is 30s, which allows for the
+     * user to be presented with a disambiguation dialog, select a wallet, and
+     * for the wallet app to subsequently start.
      */
-    retryDelayMs: 150,
+    retryDelayMs: [150, 150, 200, 500, 500, 750, 750, 1000],
```
sdlaver, line 27: @steveluscher I observed some interesting behavior here. With this backoff schedule, I observed the following connection attempts (I made a mistake in my log statement format, but the timestamps are still accurate):
Screen Shot 2022-07-23 at 2 08 58 PM

It looks like maybe the web browser is limiting the rate at which the web dapp is allowed to wake up after a timeout occurs. The first few wakeups are roughly on schedule, but after that, I see a sequence of a pretty consistent ~2s wakeup period (twice what was requested), and towards the end 3-6s.

The total wait time for this particular session was 1m14s (vs an expected ~30s).

I wouldn't want to encode this behavior into the web dapps, since it may not be consistent with browser versions, operating modes, etc. However, I'm also not sure what the UX of this extended timeout would be on web dapps. Would love your thoughts here.

steveluscher: Totally understand. I'll change this to target a deadline rather than a number of iterations.

steveluscher:
Augmented the JS implementation to wait for an absolute timeout, rather than a number of attempts.

steveluscher (approved):

PR #168, opened 2022-07-26
Add React Native bridge method to query wallet availability

sdlaver:

PR #169, opened 2022-07-26
clientlib: add a method to query wallet endpoint availability

### LINKED ISSUE #152: Add a clientlib API to detect if there are any MWA-supporting wallets present

sdlaver:
Fixes #152

### DIFF
--- a/android/clientlib/src/main/AndroidManifest.xml
+++ b/android/clientlib/src/main/AndroidManifest.xml
@@ -8,4 +8,13 @@
     <uses-permission android:name="android.permission.INTERNET" />
     <uses-permission android:name="android.permission.ACCESS_NETWORK_STATE" />
 
+    <queries>
+        <!-- Support querying whether an app with the wallet endpoint role is available -->
+        <intent>
+            <action android:name="android.intent.action.VIEW" />
+            <category android:name="android.intent.category.BROWSABLE" />
+            <data android:scheme="solana-wallet" />
+        </intent>
+    </queries>
+
 </manifest>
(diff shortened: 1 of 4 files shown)


sdlaver:
@creativedrewy not sure if we want helper library wrappers for this method - it's so straightforward it might be worth just having consumers call it directly from clientlib

@steveluscher I started on the matching changes for RN. I'm embarrassed with how quickly I didn't know the syntax conventions, so I just stopped 🥸: https://github.com/solana-mobile/mobile-wallet-adapter/pull/168

creativedrewy (approved):
Love it! If we want to make a convenience library, we can do it on top of this functionality. But I would say it isn't required for now.

steveluscher (changes requested):
Ooh. I don't _love_ this. Here's a situation we're _guaranteed_ to get into:

1. You start a dApp
2. It tells you no wallet is available
3. You leave the dApp and go install one
4. You return to the dApp expecting the ‘no wallet available prompt’ to begone

If the dApp doesn't revalidate wallet availability _every time_ you foreground it, the UI will still claim that no wallet is not installed when in fact, one is.

I've seen this pattern before on regular `@solana/wallet-adapter` and it resulted in one of the most regrettable architecture decisions in that library: infinite wallet availability polling.

Counterproposals:

1. Leave things the way they are. At the moment that someone goes to take a wallet-y action, throw a wallet-not-found type error so the dApp can show a ‘go install a wallet’ prompt.
2. Offer both a one-shot method as in this PR _and_ an event emitter that emits a `walletAvailabilityChange` event, the payload of which is a boolean.

creativedrewy:
@steveluscher It would be pretty easy in the sample Android apps to re-query on lifecycle methods. That's pretty common to do in Android land.

sdlaver:
> an event emitter that emits a `walletAvailabilityChange` event

Android doesn't let background processes watch for app package installation/removal. While an app is in the background, it's not able to wake up on this change in system state, and so can't emit or process a relevant synthetic event, such as wallet availability changes.

The only option for a standard app is to check wallet availability each time the app comes to the foreground. It's polling, but at least it's event-driven and not time-driven 🤷‍♂️.

creativedrewy:
@steveluscher @sdlaver yeah, the "polling" method is what I was referring to. In the activity, when `onStart` is called, do the intent query. Very common thing to do in Android apps.

steveluscher (approved):

PR #170, opened 2022-07-26
Make https scheme mandatory for endpoint-specific URIs

### LINKED ISSUE #58: Clarification: How much of ‘base uri’ should be honored?
Regarding the ‘[Endpoint-specific URIs](https://github.com/solana-labs/mirror-mobile-wallet-adapter/blob/main/spec/spec.md#endpoint-specific-uris)’ section of the spec.

* Should the _scheme_ that the wallet responds with be honored, or should we always use `https`?
* Should the _query_ parameters the wallet responds with be honored, or should we strip them?
* Should the _fragment_ the wallet responds with be honored, or should we strip it?

In other words, if the association path for the protocol is `/v1/associate/local?association=<association_token>&port=<port_number>` what URL should we construct if the wallet responds with:

1. `instagram://wallet`
2. `https://mydapp.com/mobilewallet/?utm_source=foo`
3. `https://mydapp.com/mobilewallet/?utm_source=foo&port=4321`
4. `https://mydapp.com/mobilewallet/#user=abc`

sdlaver:
Fixes #58

### DIFF
--- a/js/packages/mobile-wallet-adapter-protocol/src/transact.ts
+++ b/js/packages/mobile-wallet-adapter-protocol/src/transact.ts
@@ -48,6 +48,24 @@ function assertSecureContext() {
     }
 }
 
+function assertSecureEndpointSpecificURI(walletUriBase: string) {
+    let url: URL;
+    try {
+        url = new URL(walletUriBase);
+    } catch {
+        throw new SolanaMobileWalletAdapterError(
+            SolanaMobileWalletAdapterErrorCode.ERROR_FORBIDDEN_WALLET_BASE_URL,
+            'Invalid base URL supplied by wallet',
+        );
+    }
+    if (url.protocol !== 'https:') {
+        throw new SolanaMobileWalletAdapterError(
+            SolanaMobileWalletAdapterErrorCode.ERROR_FORBIDDEN_WALLET_BASE_URL,
+            'Base URLs supplied by wallets must be valid `https` URLs',
+        );
+    }
+}
+
 function getSequenceNumberFromByteArray(byteArray: ArrayBuffer): number {
     const view = new DataView(byteArray);
     return view.getUint32(0, /* littleEndian */ false);
@@ -175,7 +193,29 @@ export async function transact<TReturn>(
                                         ),
                                     );
                                     return new Promise((resolve, reject) => {
-                                        jsonRpcResponsePromises[id] = { resolve, reject };
+                                        jsonRpcResponsePromises[id] = {
+                                            resolve(result) {
+                                                switch (p) {
+                                                    case 'authorize':
+                                                    case 'reauthorize': {
+                                                        const { wallet_uri_base } = result as Awaited<
+                                                            ReturnType<MobileWallet['authorize' | 'reauthorize']>
+                                                        >;
+                                                        if (wallet_uri_base != null) {
+                                                            try {
+                                                                assertSecureEndpointSpecificURI(wallet_uri_base);
+                                                            } catch (e) {
+                                                                reject(e);
+                                                                return;
+                                                            }
+                                                        }
+                                                        break;
+                                                    }
+                                                }
+                                                resolve(result);
+                                            },
+                                            reject,
+                                        };
                                     });
                                 } as MobileWallet[TMethodName];
                             }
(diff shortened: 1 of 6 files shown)


js/packages/mobile-wallet-adapter-protocol-web3js/src/transact.ts
```diff
@@ -47,6 +49,24 @@ export async function transact<TReturn>(
             get<TMethodName extends keyof Web3MobileWallet>(target: Web3MobileWallet, p: TMethodName) {
                 if (target[p] == null) {
                     switch (p) {
+                        case 'authorize':
```
sdlaver, line 52: @steveluscher this is the most complex TS I've ever written. Did I do the thing right?

sdlaver:
I tested this change with all of:
- fakedapp
- example web app
- example React Native app

js/packages/mobile-wallet-adapter-protocol-web3js/src/transact.ts
```diff
@@ -158,3 +178,18 @@ export async function transact<TReturn>(
     };
     return await baseTransact(augmentedCallback, config);
 }
+
+function validateSecureEndpointSpecificURI(walletUriBase?: string) {
```
steveluscher, line 182: Nullability will kill us all. Prefer: make this argument required, then just don't call this method if you don't have a base URI.

js/packages/mobile-wallet-adapter-protocol-web3js/src/transact.ts
```diff
@@ -158,3 +178,18 @@ export async function transact<TReturn>(
...
+    if (walletUriBase) {
+        let url: URL | null = null;
+        try {
+            url = new URL(walletUriBase);
```
steveluscher, line 186: Fun fact: React Native doesn't offer the `URL` class. You have to polyfill it. Annoying.

js/packages/mobile-wallet-adapter-protocol-web3js/src/transact.ts
```diff
@@ -158,3 +178,18 @@ export async function transact<TReturn>(
...
+        } catch {} // eslint-disable-line no-empty
```
steveluscher, line 187: If we're saying that we can't let through non-https URLs we probably also shouldn't let through URLs that straight up don't parse.

js/packages/mobile-wallet-adapter-protocol-web3js/src/transact.ts
sdlaver: Done

js/packages/mobile-wallet-adapter-protocol-web3js/src/transact.ts
sdlaver: Done

sdlaver:
> Move all the JS code into `mobile-wallet-adapter-protocol/src/transact.ts` and the web3 version will get it for free!

I looked at the implementation of `mobile-wallet-adapter-protocol/src/transact.ts` and don't feel like I have a sufficient grasp on TS syntax to do this correctly. Would you mind demonstrating?

steveluscher:
Oof. You're right. That was kind of gnarly. Commit added. Ready to land.

js/packages/mobile-wallet-adapter-protocol/src/transact.ts
```diff
@@ -175,7 +193,29 @@ export async function transact<TReturn>(
...
+                                            resolve(result) {
+                                                switch (p) {
+                                                    case 'authorize':
+                                                    case 'reauthorize': {
+                                                        const { wallet_uri_base } = result as Awaited<
+                                                            ReturnType<MobileWallet['authorize' | 'reauthorize']>
+                                                        >;
+                                                        if (wallet_uri_base != null) {
+                                                            try {
+                                                                assertSecureEndpointSpecificURI('http://apple.com');
```
sdlaver, line 206: Cool, but test code, I assume?

steveluscher (approved):

js/packages/mobile-wallet-adapter-protocol/src/transact.ts
steveluscher: No. If that throws, we need to catch it and send it through `reject()` so that the promise for `wallet.authorize()` rejects with the thrown error. Otherwise the whole system halts because the `authorize` promise never resolves _nor_ rejects.

steveluscher: OH. Haha.

* scrolls over *

Notices the URL.

sdlaver:
I will plan to merge this tomorrow. Since it's a spec change, I'd like to give 24 hours after posting into the Discord #wallet channel for any feedback.

PR #171, opened 2022-07-26
feat: you can now configure the `SolanaMobileWalletAdapter` with an address selector

steveluscher:
# Background

Mobile wallet adapter allows us to authorize multiple addresses from a single click of the ‘connect’ button. This causes an impedance mismatch between MWA and `@solana/wallet-adapter` that doesn't have the concept of multiple addresses.

# Problem statement

The existing `wallet-adapter` API is single-key, like this:

```typescript
const {publicKey, signTransaction} = useWallet();
```

How, in this world, should `wallet-adapter` make the choice of `publicKey`, if the user decides to authorize multiple addresses in a mobile wallet adapter session?

# Summary of changes

* Added a configuration param to `SolanaMobileWalletAdapter` called `addressSelector`. Developers can provide an implementation of this to make a sensible selection of address given multiple addresses.
* Made a `createDefaultAddressSelector` helper, that just chooses the first address in the list.

Note, that because the selector is async, app developers have a great amount of freedom to pause execution of the wallet adapter plugin to show – for instance – an address selection UI to their users.

### DIFF
--- a/js/packages/wallet-adapter-mobile/src/adapter.ts
+++ b/js/packages/wallet-adapter-mobile/src/adapter.ts
@@ -1,10 +1,5 @@
 import { Web3MobileWallet, transact } from '@solana-mobile/mobile-wallet-adapter-protocol-web3js';
-import {
-    AppIdentity,
-    AuthorizationResult,
-    AuthToken,
-    Base64EncodedAddress,
-} from '@solana-mobile/mobile-wallet-adapter-protocol';
+import { AppIdentity, AuthorizationResult, Base64EncodedAddress } from '@solana-mobile/mobile-wallet-adapter-protocol';
 import {
     BaseMessageSignerWalletAdapter,
     WalletConnectionError,
@@ -29,6 +24,10 @@ export interface AuthorizationResultCache {
     set(authorizationResult: AuthorizationResult): Promise<void>;
 }
 
+export interface AddressSelector {
+    select(addresses: Base64EncodedAddress[]): Promise<Base64EncodedAddress>;
+}
+
 export const SolanaMobileWalletAdapterWalletName = 'Default wallet app' as WalletName;
 
 const SIGNATURE_LENGTH_IN_BYTES = 64;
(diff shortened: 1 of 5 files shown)


js/packages/wallet-adapter-mobile/src/adapter.ts
```diff
@@ -74,12 +77,9 @@ export class SolanaMobileWalletAdapter extends BaseMessageSignerWalletAdapter {
...
 
     get publicKey(): PublicKey | null {
-        if (this._publicKey == null && this._authorizationResult != null) {
+        if (this._publicKey == null && this._selectedAddress != null) {
             try {
-                this._publicKey = getPublicKeyFromAddress(
-                    // TODO(#44): support multiple addresses
-                    this._authorizationResult.addresses[0],
-                );
+                this._publicKey = getPublicKeyFromAddress(this._selectedAddress);
```
jordansexton, line 82: I might be missing something here, but how can you reliably get a public key from an address? Addresses (if not on Solana) may not be reversible to pubkeys.

js/packages/wallet-adapter-mobile/src/adapter.ts
```diff
@@ -29,6 +24,10 @@ export interface AuthorizationResultCache {
     set(authorizationResult: AuthorizationResult): Promise<void>;
 }
 
+export interface AddressSelector {
+    select(addresses: AuthorizationResult['addresses']): Promise<AuthorizationResult['addresses'][number]>;
```
jordansexton, line 28: Maybe a stylistic choice since the resolved type will be the same, but `AuthorizationResult['addresses']`, and definitely `AuthorizationResult['addresses'][number]`, are less clear than their `Base64EncodedAddress[] / Base64EncodedAddress`.

steveluscher: Good question. I think there are two positions you can take here:

1. This is a thing that turns `T[]` into `T`
2. This is a thing that takes `ThisParticularType[]` and turns it into `ThisParticularType`

My goal with making it a lookup on `addresses` was to make Typescript do more work for me; if the type changes then that change automatically flows through this function.

My implementation actually hits somewhere in between option 1 and 2, and maybe I should go for option 1 which just says ‘I’m going to give you a list of who knows what and you give me a single item back.’

js/packages/wallet-adapter-mobile/src/adapter.ts
steveluscher: I guess I presumed that the Solana `wallet-adapter` would only ever work with Solana wallets. We’d have to change the return type of `useWallet`, which right now only has the concept of `publicKey`.

jordansexton: Ah, that's a fair assumption then! I think I was not seeing that the context of this code is `wallet-adapter-mobile` and not `mobile-wallet-adapter`

js/packages/wallet-adapter-mobile/src/adapter.ts
jordansexton: Oh, you like writing Java? Name every `AsyncSelector { select<T>(items: T[]): Promise<T> }`

steveluscher: I like not writing types at all and making Typescript work for the money I spent on it.

Uh… wait.

sdlaver (commented):
Better late than never. Approved!

Issue #44, opened 2022-06-14
Add multi-pubkey return support for the authorize method


sdlaver:
**Spec**
- [ ] Update `sign_and_send_transactions`, `sign_transactions`, and `sign_messages` to accept an `account` param.

**Fakewallet**
- [ ] Allow user to choose addresses to share with the dApp during `authorize`

**JavaScript protocol**
- [x] Augment `signMessages` to accept `addresses` param
  https://github.com/solana-mobile/mobile-wallet-adapter/pull/176
* ~Augment `signTransactions` to accept `address` param~ Now inferred from transaction.
* ~Augment `signAndSendTransactions` to accept `address` param~ Now inferred from transaction.

**React Native Example App**
- [X] Show dropdown selector when more than one pubkey
       #173
- [x] Update calls to protocol API to specify the address that should sign
   https://github.com/solana-mobile/mobile-wallet-adapter/pull/176

**JavaScript `wallet-adapter-mobile` plugin**
- [x] Offer way for dApps to inject an address chooser method
        #171

**Mobile web example app**
- [ ] Use address chooser to present an address selection UI to users

PR #173, opened 2022-07-27
Offer menu to select address when more than one are authorized (React Native app)

steveluscher:
When the `authorize` or `reauthorize` methods return more than one address, offer a menu to switch the active address.

https://user-images.githubusercontent.com/13243/181186391-8cfde9e2-8c66-4315-a9e7-d6c1bcc06def.mov

Addresses #44.


steveluscher:
Ignore the first two commits; they're from #176.

sdlaver (approved):

Issue #175, opened 2022-07-27
The sign_messages method should accept an address parameter


sdlaver:
This aligns with the spec changes from #111

PR #176, opened 2022-07-27
Thread the currently selected address through to each `signMessages` call

steveluscher:
#### Problem

Unlike transactions, from which the requisite signer can be _inferred_, when we call methods like `signMessages` we have to _tell_ the wallet which authorized address we would like to sign the message with.

#### Summary of changes

* Added an `address` param to the `signMessages` API of the JavaScript protocol implementation.
* Updated the JavaScript example apps to supply `account` to `signMessages()`.

Addresses #175.

### DIFF
--- a/examples/example-react-native-app/utils/useAuthorization.tsx
+++ b/examples/example-react-native-app/utils/useAuthorization.tsx
@@ -11,24 +11,30 @@ import {toUint8Array} from 'js-base64';
 import {useCallback} from 'react';
 import useSWR from 'swr';
 
+type Account = Readonly<{
+  address: Base64EncodedAddress;
+  publicKey: PublicKey;
+}>;
+
 const STORAGE_KEY = 'cachedAuthorization';
 
 function getDataFromAuthorizationResult(
   authorizationResult: AuthorizationResult,
 ) {
   return {
+    account: getAccountFromAuthorizationResult(authorizationResult),
     authorization: authorizationResult,
-    publicKey: getPublicKeyFromAuthorizationResult(authorizationResult),
   };
 }
 
-function getPublicKeyFromAuthorizationResult(
+function getAccountFromAuthorizationResult(
   authorizationResult: AuthorizationResult,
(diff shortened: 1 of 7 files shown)


js/packages/wallet-adapter-mobile/src/adapter.ts
```diff
@@ -190,6 +192,7 @@ export class SolanaMobileWalletAdapter extends BaseMessageSignerWalletAdapter {
         this._authorizationResultCache.clear(); // TODO: Evaluate whether there's any threat to not `awaiting` this expression
         delete this._authorizationResult;
         delete this._publicKey;
+        delete this._selectedAddress;
```
steveluscher, line 195: Sneaking this unrelated change in here.

examples/example-react-native-app/utils/useAuthorization.tsx
```diff
@@ -11,24 +11,30 @@ import {toUint8Array} from 'js-base64';
 import {useCallback} from 'react';
 import useSWR from 'swr';
 
+type Account = Readonly<{
```
sdlaver, line 14: Can you elaborate on the thinking behind this type? From a chain-agnostic POV, it makes sense; addresses and public keys aren't necessarily the same. However, this is in a Solana network example, where we can safely make that assumption. I'm not opposed to it, since it does encode a valuable distinction in an example that is likely to be copied and reused, just curious to hear your thinking.

steveluscher: For sure!

At some point, this app is going to need to call `signMessage` and when it does, the API will demand that it specifies an `account`. If I throw away the base64 encoded account string, I’ll need to reconstruct it from the pubkey later.

Secondly: at some point the user will switch accounts using a combo box. If I’ve thrown away the base64 encoded account strings, I’ll have to reconstruct the selected one from the pubkey in order to make the choice of account to switch to.

sdlaver (approved):

PR #177, opened 2022-07-28
Initial implementation of client-ktx sample app

creativedrewy:
[kotlin.webm](https://user-images.githubusercontent.com/2018851/181635816-2320c0b0-7434-4a7b-8e83-d6f7b7088508.webm)

### DIFF
--- a/examples/example-clientlib-ktx-app/app/src/main/java/com/solanamobile/ktxclientsample/viewmodel/SampleViewModel.kt
+++ b/examples/example-clientlib-ktx-app/app/src/main/java/com/solanamobile/ktxclientsample/viewmodel/SampleViewModel.kt
@@ -1,16 +1,12 @@
 package com.solanamobile.ktxclientsample.viewmodel
 
 import android.net.Uri
-import android.util.Log
 import androidx.lifecycle.ViewModel
 import androidx.lifecycle.viewModelScope
-import com.solana.Solana
-import com.solana.api.requestAirdrop
 import com.solana.core.PublicKey
 import com.solana.mobilewalletadapter.clientlib.ActivityResultSender
 import com.solana.mobilewalletadapter.clientlib.MobileWalletAdapter
-import com.solana.networking.NetworkingRouter
-import com.solana.networking.RPCEndpoint
+import com.solanamobile.ktxclientsample.usecase.SolanaRpcUseCase
 import dagger.hilt.android.lifecycle.HiltViewModel
 import kotlinx.coroutines.flow.MutableStateFlow
 import kotlinx.coroutines.flow.StateFlow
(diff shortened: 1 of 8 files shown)


examples/example-clientlib-ktx-app/app/src/main/java/com/solanamobile/ktxclientsample/viewmodel/SampleViewModel.kt
```diff
@@ -3,41 +3,172 @@ package com.solanamobile.ktxclientsample.viewmodel
...
+        val connection = persistanceUseCase.getWalletConnection()
+        if (connection is Connected) {
+            _state.value.copy(
+                isLoading = true,
+                canTransact = true,
+                userAddress = connection.publickKey.toBase58()
+            ).updateViewState()
+
+            viewModelScope.launch {
+                val balance = solanaRpcUseCase.getBalance(connection.publickKey)
```
creativedrewy, line 64: Decided against persisting sol balance, favoring querying it live.

examples/example-clientlib-ktx-app/app/src/main/java/com/solanamobile/ktxclientsample/viewmodel/SampleViewModel.kt
```diff
@@ -3,41 +3,172 @@ package com.solanamobile.ktxclientsample.viewmodel
...
+    }
+
+    fun addFunds(sender: ActivityResultSender) {
         viewModelScope.launch {
-            walletAdapter.transact(sender) {
-                val result = authorize(Uri.parse("https://solana.com"), Uri.parse("favicon.ico"), "Solana")
+            val conn = persistanceUseCase.getWalletConnection()
+
+            val currentConn = walletAdapter.transact(sender) {
+                when (conn) {
```
creativedrewy, line 79: Not handled by either of these cases: when the wallet tells us it's time to authorize fresh.

examples/example-clientlib-ktx-app/app/src/main/java/com/solanamobile/ktxclientsample/viewmodel/SampleViewModel.kt
```diff
@@ -3,41 +3,172 @@ package com.solanamobile.ktxclientsample.viewmodel
...
+
+                val tx = Transaction()
+                tx.add(MemoProgram.writeUtf8(conn.publickKey, memoText))
+                tx.setRecentBlockHash(blockHash!!)
+                tx.feePayer = conn.publickKey
+
+                val bytes = tx.serialize(SerializeConfig(requireAllSignatures = false))
+
+                val result = walletAdapter.transact(sender) {
+                    reauthorize(solanaUri, iconUri, identityName, conn.authToken)
```
creativedrewy, line 139: The user won't be able to select this option until after they have authorized, so that's why we are assuming reauth. But like above, what's not handled is if the wallet tells us the token is invalid. Now I'm wondering too if I need to re-persist what is returned from reauthorize...

PR #183, opened 2022-07-29
sign_messages should accept multiple addresses for signing

### LINKED ISSUE #175: The sign_messages method should accept an address parameter

### LINKED ISSUE #178: sign_messages should allow signing with multiple addresses (parity of behavior with sign_transactions)

sdlaver:
Fixes #175
Fixes #178

### DIFF
--- a/examples/example-react-native-app/components/SignMessageButton.tsx
+++ b/examples/example-react-native-app/components/SignMessageButton.tsx
@@ -27,11 +27,12 @@ export default function SignMessageButton({children, message}: Props) {
       const [signature] = await transact(async wallet => {
         const freshAccount = await authorizeSession(wallet);
         return await wallet.signMessages({
-          address:
+          addresses: [
             // Either the address that was already selected when this method was called...
             selectedAccount?.address ??
             // ...or the newly authorized address
-            freshAccount.address,
+            freshAccount.address
+          ],
           payloads: [buffer],
         });
       });
--- a/js/packages/mobile-wallet-adapter-protocol-web3js/src/transact.ts
+++ b/js/packages/mobile-wallet-adapter-protocol-web3js/src/transact.ts
@@ -27,7 +27,7 @@ interface Web3SignTransactionsAPI {
 }
 
 interface Web3SignMessagesAPI {
-    signMessages(params: { address: Base64EncodedAddress; payloads: Uint8Array[] }): Promise<Uint8Array[]>;
+    signMessages(params: { addresses: Base64EncodedAddress[]; payloads: Uint8Array[] }): Promise<Uint8Array[]>;
 }
 
 export interface Web3MobileWallet
--- a/js/packages/mobile-wallet-adapter-protocol/src/types.ts
+++ b/js/packages/mobile-wallet-adapter-protocol/src/types.ts
@@ -63,7 +63,7 @@ export interface ReauthorizeAPI {
 }
 export interface SignMessagesAPI {
     signMessages(params: {
-        address: Base64EncodedAddress;
+        addresses: Base64EncodedAddress[];
         payloads: Base64EncodedMessage[];
     }): Promise<Readonly<{ signed_payloads: Base64EncodedSignedMessage[] }>>;
 }
--- a/js/packages/wallet-adapter-mobile/src/adapter.ts
+++ b/js/packages/wallet-adapter-mobile/src/adapter.ts
@@ -269,7 +269,7 @@ export class SolanaMobileWalletAdapter extends BaseMessageSignerWalletAdapter {
                 return await this.transact(async (wallet) => {
                     await this.performReauthorization(wallet, authToken);
                     const [signedMessage] = await wallet.signMessages({
-                        address: selectedAddress,
+                        addresses: [selectedAddress],
                         payloads: [message],
                     });
                     const signature = signedMessage.slice(-SIGNATURE_LENGTH_IN_BYTES);
--- a/android/clientlib-ktx/src/main/java/com/solana/mobilewalletadapter/clientlib/AdapterOperations.kt
+++ b/android/clientlib-ktx/src/main/java/com/solana/mobilewalletadapter/clientlib/AdapterOperations.kt
@@ -8,7 +8,7 @@ interface AdapterOperations {
     suspend fun reauthorize(identityUri: Uri, iconUri: Uri, identityName: String, authToken: String): MobileWalletAdapterClient.AuthorizationResult
     suspend fun deauthorize(authToken: String)
     suspend fun getCapabilities(): MobileWalletAdapterClient.GetCapabilitiesResult
-    suspend fun signMessages(transactions: Array<ByteArray>): MobileWalletAdapterClient.SignPayloadsResult
+    suspend fun signMessages(messages: Array<ByteArray>, addresses: Array<ByteArray>): MobileWalletAdapterClient.SignPayloadsResult
     suspend fun signTransactions(transactions: Array<ByteArray>): MobileWalletAdapterClient.SignPayloadsResult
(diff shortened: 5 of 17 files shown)


sdlaver:
@steveluscher some minor TS updates in this PR

js/packages/wallet-adapter-mobile/src/adapter.ts
```diff
@@ -269,7 +269,7 @@ export class SolanaMobileWalletAdapter extends BaseMessageSignerWalletAdapter {
                 return await this.transact(async (wallet) => {
                     await this.performReauthorization(wallet, authToken);
                     const [signedMessage] = await wallet.signMessages({
-                        address: selectedAddress,
+                        addresses: [selectedAddress],
                         payloads: [message],
                     });
```
steveluscher, line 274: This now returns a list of a list of signed messages, right?

```typescript
const [[signedMessage]] = await /* ... */
```

js/packages/mobile-wallet-adapter-protocol-web3js/src/transact.ts
```diff
@@ -27,7 +27,7 @@ interface Web3SignTransactionsAPI {
 }
 
 interface Web3SignMessagesAPI {
-    signMessages(params: { address: Base64EncodedAddress; payloads: Uint8Array[] }): Promise<Uint8Array[]>;
+    signMessages(params: { addresses: Base64EncodedAddress[]; payloads: Uint8Array[] }): Promise<Uint8Array[]>;
```
steveluscher, line 30: Now this has to return a list of signature arrays, right?

```typescript
Promise<Uint8Array[][]>;
```

js/packages/mobile-wallet-adapter-protocol/src/types.ts
```diff
@@ -63,7 +63,7 @@ export interface ReauthorizeAPI {
 }
 export interface SignMessagesAPI {
     signMessages(params: {
-        address: Base64EncodedAddress;
+        addresses: Base64EncodedAddress[];
         payloads: Base64EncodedMessage[];
     }): Promise<Readonly<{ signed_payloads: Base64EncodedSignedMessage[] }>>;
```
steveluscher, line 68: Now this has to return a list of a list of signed messages, right?

```typescript
Promise<Readonly<{ signed_payloads: Base64EncodedSignedMessage[][] }>>;
```

examples/example-react-native-app/components/SignMessageButton.tsx
```diff
@@ -27,11 +27,12 @@ export default function SignMessageButton({children, message}: Props) {
       const [signature] = await transact(async wallet => {
```
steveluscher, line 27: This now returns a list of a list of signatures, right?

```typescript
const [[signature]] = await transact(async wallet => {
```

steveluscher (changes requested):
I imagine that `SignPayloadsResult` probably suffers the same type mismatch now, where it needs to return an array of an array of signed payloads.

I don't know if we should consider going as far as to change the API to return:

```
type Base64SignedMessage = string;
type SignedPayloadsForAddresses = Base64SignedMessage[];
type SignResult = Readonly<{
  signed_payloads_for_addresses: SignedPayloadsForAddresses[],
}>;
```

js/packages/mobile-wallet-adapter-protocol-web3js/src/transact.ts
sdlaver: `Uint8Array[]`

TS syntax question: is this an array of uint8, or an array of uint8array?

sdlaver: If it's the latter, then this returns an array of signed messages (where each message may have several signatures appended)

steveluscher: `Array<Uint8Array>` and `Uint8Array[]` are synonyms. Those are arrays of `Uint8Array` instances.

Still, when we add multi-address signing, we need to end up in this scenario:

```
const foo = signMessages({addresses: [A,B,C], payloads: [Uint8Array([1]), Uint8Array([2]), Uint8Array([3])]);
console.log(foo);
```

```
[
  [signatureOfPayload1ByAddressA, signatureOfPayload1ByAddressB, signatureOfPayload1ByAddressC],
  [signatureOfPayload2ByAddressA, signatureOfPayload2ByAddressB, signatureOfPayload2ByAddressC],
  [signatureOfPayload3ByAddressA, signatureOfPayload3ByAddressB, signatureOfPayload3ByAddressC],
]
```

That's a `Uint8Array[][]` or synonymously an `Array<Array<Uint8Array>>`.

steveluscher: This is also just an opinion. You could equally argue for:

```
[
  [signatureOfPayload1ByAddressA, signatureOfPayload2ByAddressA, signatureOfPayload3ByAddressA],
  [signatureOfPayload1ByAddressB, signatureOfPayload2ByAddressB, signatureOfPayload3ByAddressB],
  [signatureOfPayload1ByAddressC, signatureOfPayload2ByAddressC, signatureOfPayload3ByAddressC],
]
```

But still the type is `UintArray[][]`.

sdlaver: Ah, I see now. This RPC method does not return signatures though; it returns the signed payload. This would be [`message1 || sigA || sigB || ... || sigN`, `message2 || sigA || sigB || ... || sigN`, ...]

steveluscher: Alternative API: `Readonly<{[address: Base64EncodedAddress]: Uint8Array[]}>`

```
{
  A: [payload1SignatureByAddressA, payload2SignatureByAddressA, payload3SignatureByAddressA],
  B: [payload1SignatureByAddressB, payload2SignatureByAddressB, payload3SignatureByAddressB],
  C: [payload1SignatureByAddressC, payload2SignatureByAddressC, payload3SignatureByAddressC],
}
```

sdlaver: We had spec'ed it out the current way to allow for the wallets and/or chain-specific extensions to specify how signatures are attached to messages. For eg, the message may need to be modified to add a domain prefix, wrapped in a PKCS#7 wrapper, etc.

jordansexton: > This RPC method does not return signatures though; it returns the signed payload.

Ah, good point. This keeps the return types a bit more consistent with signed transactions.

steveluscher: > This RPC method does not return signatures though; it returns the signed payload

Right right. Getting it mixed up with `@solana/wallet-adapter` which does return just the signature.

steveluscher: Oof. OK, I remember having a conversation with one of you that went something like ‘does `signMessages` return `sign(sign(message, A), B)` or `[sign(message, A), sign(message, B)]` and I recall the answer was the latter.

¯\_(°ペ)_/¯

steveluscher (approved):

js/packages/mobile-wallet-adapter-protocol-web3js/src/transact.ts
jordansexton: `signMessage` and `signTransaction` work the same way -- all signers sign the same payload, and have their signatures added to the return. Each signature is independent.

steveluscher: I think the problem with this conversation thread is that when you say ‘`signMessages` does…’ you have to specify whether you mean `wallet-adapter` or `mobile-wallet-adapter-protocol` because right now those two do different things.

PR #184, opened 2022-07-30
Update spec and JavaScript impl to add human-readable labels to accounts

steveluscher:
# Problem

Using a dApp today demands that you memorize part of a base58 encoded public key, because that's the primary way by which a given account in your wallet is represented.

In the _wallets_ on the other hand, you're allowed to add convenient, human-readable labels, like ‘My NFT wallet.’

# Summary of changes

- [X] Update the spec to return `{address: Base64EncodedAddress, label?: string}` from `authorize` and `reauthorize`. dApps can now use that label in their UI for a better experience.
- [x] Update server/client implementation to supply labels with accounts
- [x] Update JavaScript protocol method signatures
- [x] Update React Native Example App to display labels
- [x] Update Web example app to handle accounts with labels

# Test plan

React Native example app, with address label:

image

Addresses #187.

### DIFF
--- a/android/walletlib/src/main/java/com/solana/mobilewalletadapter/walletlib/protocol/MobileWalletAdapterServer.java
+++ b/android/walletlib/src/main/java/com/solana/mobilewalletadapter/walletlib/protocol/MobileWalletAdapterServer.java
@@ -146,9 +146,14 @@ public class MobileWalletAdapterServer extends JsonRpc20Server {
             final JSONObject o = new JSONObject();
             try {
                 o.put(ProtocolContract.RESULT_AUTH_TOKEN, result.authToken);
-                final JSONArray addresses = new JSONArray(); // TODO(#44): support multiple addresses
-                addresses.put(publicKeyBase64);
-                o.put(ProtocolContract.RESULT_ADDRESSES, addresses);
+                final JSONArray accounts = new JSONArray().put(
+                        // TODO(#44): support multiple accounts
+                        new JSONObject()
+                                .put(ProtocolContract.RESULT_ACCOUNTS_ADDRESS, publicKeyBase64)
+                                // TODO(#187): Implement account labels
+                                .put(ProtocolContract.RESULT_ACCOUNTS_LABEL, "My First Wallet")
+                );
+                o.put(ProtocolContract.RESULT_ACCOUNTS, accounts);
                 o.put(ProtocolContract.RESULT_WALLET_URI_BASE, result.walletUriBase); // OK if null
             } catch (JSONException e) {
                 throw new RuntimeException("Failed preparing authorization response", e);
--- a/spec/spec.md
+++ b/spec/spec.md
@@ -328,7 +328,10 @@ where:
 ```
 {
     “auth_token”: “<auth_token>”,
-    “addresses”: [“<address>", ...]
+    “accounts”: [
+        {“address”: “<address>", “label”: “<label>”},
+        ...
+    ],
     “wallet_uri_base”: “<wallet_uri_base>”,
 }
 ```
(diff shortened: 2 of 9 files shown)


steveluscher:
Before going much further on this, I'll just make the spec updates so that we can talk about those.

spec/spec.md
```diff
@@ -328,15 +328,20 @@ where:
...
 }
 ```
 
 where:
 
 - `auth_token`: an opaque string representing a unique identifying token issued by the wallet endpoint to the dapp endpoint. The format and contents are an implementation detail of the wallet endpoint. The dapp endpoint can use this on future connections to `reauthorize` access to [privileged methods](#privileged-methods).
-- `addresses`: one or more base64-encoded addresses, for the accounts to which this auth token corresponds
+- `labelled_addresses`: one or more value objects that represent the accounts to which this auth token corresponds. These objects hold the following properties:
+  - `address`: a base64-encoded address
+  - `label`: (optional) a human-readable string that describes the address. Wallets that allow their users to label their accounts may choose to return those labels here to enhance the user experience in the target dapp.
```
sdlaver, line 344: NIT: I've been using the terms "wallet endpoints" and "dapp endpoints", rather than just "wallet" and "dapp", to avoid confusing wallet-applications with wallet-accounts

spec/spec.md
```diff
@@ -328,15 +328,20 @@ where:
...
     “wallet_uri_base”: “<wallet_uri_base>”,
 }
 ```
 
 where:
 
 - `auth_token`: an opaque string representing a unique identifying token issued by the wallet endpoint to the dapp endpoint. The format and contents are an implementation detail of the wallet endpoint. The dapp endpoint can use this on future connections to `reauthorize` access to [privileged methods](#privileged-methods).
-- `addresses`: one or more base64-encoded addresses, for the accounts to which this auth token corresponds
+- `labelled_addresses`: one or more value objects that represent the accounts to which this auth token corresponds. These objects hold the following properties:
+  - `address`: a base64-encoded address
```
sdlaver, line 343: "... for this account"

spec/spec.md
```diff
@@ -328,15 +328,20 @@ where:
 ```
 {
     “auth_token”: “<auth_token>”,
-    “addresses”: [“<address>", ...]
+    “labelled_addresses”: [
```
sdlaver, line 331: WDYT about naming this field `accounts`? Then you'd have, e.g., `accounts[0].address` and `accounts[0].label`, which captures the semantics of the array and object pretty well.

steveluscher: Definitely. I've simply given up on trying to choose the right name, in favor of word salad-ing into a PR to let you come up with better ones :)

steveluscher:
I made a change here, but I'm not sure I implemented your feedback about ‘wallet endpoint’ and ‘dapp endpoint’ correctly. The language still feels a bit stilted.

sdlaver (approved):

android/walletlib/src/main/java/com/solana/mobilewalletadapter/walletlib/protocol/MobileWalletAdapterServer.java
```diff
@@ -146,9 +146,14 @@ private void onAuthorizationComplete(@NonNull NotifyOnCompleteFuture<Authorizati
...
                 o.put(ProtocolContract.RESULT_AUTH_TOKEN, result.authToken);
-                final JSONArray addresses = new JSONArray(); // TODO(#44): support multiple addresses
-                addresses.put(publicKeyBase64);
-                o.put(ProtocolContract.RESULT_ADDRESSES, addresses);
+                final JSONArray accounts = new JSONArray().put(
+                        // TODO(#44): support multiple accounts
+                        new JSONObject()
+                                .put(ProtocolContract.RESULT_ACCOUNTS_ADDRESS, publicKeyBase64)
+                                // TODO(#187): Implement account labels
+                                .put(ProtocolContract.RESULT_ACCOUNTS_LABEL, "My First Wallet")
```
sdlaver, line 154: I'd propose that the label be empty until #187 is complete.

sdlaver (changes requested):

PR #186, opened 2022-08-01
spec: parameters for sign_and_send_transactions

### LINKED ISSUE #134: Should the sign_and_send_transaction method include a commitment parameter?
This was discussed as part of #111. We decided to unblock merging that PR by filing this issue, and continue the conversation on `commitment` separately.

sdlaver:
Fixes #134


spec/spec.md
```diff
@@ -631,36 +631,36 @@ where:
...
+      “submitted”: [<submitted>, ...],
   }
   ```
 
-  if the requested commitment level was not reached for any of the signed transactions, where:
+  if the wallet endpoint was unable to submit one or more of the signed transactions to the network, where:
 
   - `signatures`: as defined for a successful result
-  - `commitment_reached`: for each entry in `signatures`, a boolean indicating whether the desired commitment level was reached
+  - `submitted`: for each entry in `signatures`, a boolean indicating whether the transaction was submitted to the network
```
jordansexton, line 646: Feels like we should define some of this a bit more clearly --

> a successful result

What does this mean? If the signature / txid was returned by the RPC node, is that a successful result?

> submitted to the network

What does this mean? Why would a transaction _not_ be submitted to the network? Does it mean an RPC node didn't respond with the txid?

sdlaver: > > a successful result
> 
> What does this mean? If the signature / txid was returned by the RPC node, is that a successful result?

Basically, if [`sendTransaction`](https://docs.solana.com/developing/clients/jsonrpc-api#sendtransaction) returns success. I'll make the text more explicit on what a success is.

> > submitted to the network
> 
> What does this mean? Why would a transaction not be submitted to the network? Does it mean an RPC node didn't respond with the txid?

Same answer - did an RPC node accept the transaction. I'll update the text to better capture the low bar of success for this method call.

jordansexton: Hmm. So in the case that sending failed, the wallet should still respond with signatures, which the dapp could then send itself? This sounds like a potential security issue, since the `signAndSendTransaction` flow is generally preferred by wallets to avoid exposing the signature _unless_ the wallet does the sending.

jordansexton: Could responding with `{ signatures: [<signature>, null, <signature>] }` instead be an option? I'm not sure how useful or reliable the information from `submitted` is, because a dapp has to check to see if the tx was processed/confirmed/finalized anyway. If it wasn't submitted, I would think the dapp should request the wallet to sign and send it again rather than send itself. For example, it may be a metatransaction anyway, and therefore the dapp can't just send itself.

sdlaver: Yep, that sounds reasonable - a null entry in the array conveys the failure of sending a particular transaction in the set, and avoids exposing a signature that a dapp could attempt to use itself.

spec/spec.md
```diff
@@ -598,7 +598,7 @@ where:
 
 - `payloads`: one or more base64-encoded transaction payload to sign
 - `options`: a JSON object, containing:
```
steveluscher, line 600: I'm totally fine with this being non-optional, I just want to make sure that was on purpose.

spec/spec.md
```diff
@@ -598,7 +598,7 @@ where:
...
-  - `commitment`: (optional) if set, one of `processed`, `confirmed`, or `finalized`, specifying the desired commitment level that must be reached for this request to be successful. If not set, defaults to `confirmed`.
+  - `min_context_slot`: (optional) if set, the minimum slot number that the request can be evaluated at
```
steveluscher, line 601: I updated the Solana docs for this parameter, and those edits might be helpful here too.

https://github.com/solana-labs/solana/commit/a0698d4cc3ba88510cdd2b869e9c7bc8448ab86a

spec/spec.md
sdlaver: It can be optional - thanks, I'll fix.

PR #189, opened 2022-08-01
fix: move fee payer and blockhash backfilling from protocol implementation to wallet adapter plugin

steveluscher:
The web3.js protocol wrapper was only ever meant to translate between `@solana/web3.js` _datatypes_ and the raw underlying API's datatypes. Right now it does too much work.

# Summary of changes

* Move fee payer backfilling into the wallet adapter plugin
* Move blockhash backfilling into the wallet adapter plugin and choose the lower of the two commitments.

### DIFF
--- a/js/packages/mobile-wallet-adapter-protocol-web3js/src/transact.ts
+++ b/js/packages/mobile-wallet-adapter-protocol-web3js/src/transact.ts
@@ -1,10 +1,9 @@
-import { Connection, PublicKey, Transaction, TransactionSignature } from '@solana/web3.js';
+import { Transaction, TransactionSignature } from '@solana/web3.js';
 import {
     AuthorizeAPI,
     Base64EncodedAddress,
     CloneAuthorizationAPI,
     DeauthorizeAPI,
-    Finality,
     MobileWallet,
     ReauthorizeAPI,
     transact as baseTransact,
@@ -16,8 +15,6 @@ import { fromUint8Array, toUint8Array } from './base64Utils';
 
 interface Web3SignAndSendTransactionsAPI {
     signAndSendTransactions(params: {
-        connection: Connection;
-        fee_payer?: PublicKey;
         minContextSlot?: number;
         transactions: Transaction[];
     }): Promise<TransactionSignature[]>;
@@ -51,43 +48,13 @@ export async function transact<TReturn>(
                     switch (p) {
                         case 'signAndSendTransactions':
                             target[p] = async function ({
-                                connection,
-                                fee_payer: feePayer,
                                 minContextSlot,
                                 transactions,
                                 ...rest
                             }: Parameters<Web3MobileWallet['signAndSendTransactions']>[0]) {
-                                let latestBlockhashPromise: ReturnType<
-                                    InstanceType<typeof Connection>['getLatestBlockhash']
-                                >;
-                                let targetCommitment: Finality;
-                                switch (connection.commitment) {
-                                    case 'confirmed':
-                                    case 'finalized':
-                                    case 'processed':
-                                        targetCommitment = connection.commitment;
-                                        break;
-                                    default:
-                                        targetCommitment = 'finalized';
-                                }
-                                function getLatestBlockhashPromise() {
-                                    if (latestBlockhashPromise == null) {
-                                        latestBlockhashPromise = connection.getLatestBlockhash({
-                                            commitment: targetCommitment,
-                                        });
-                                    }
-                                    return latestBlockhashPromise;
-                                }
                                 const payloads = await Promise.all(
                                     transactions.map(async (transaction) => {
-                                        if (transaction.feePayer == null) {
-                                            transaction.feePayer = feePayer;
-                                        }
-                                        if (transaction.recentBlockhash == null) {
-                                            const { blockhash } = await getLatestBlockhashPromise();
-                                            transaction.recentBlockhash = blockhash;
-                                        }
-                                        const serializedTransaction = transaction.serialize({
+                                        const serializedTransaction = await transaction.serialize({
                                             requireAllSignatures: false,
                                             verifySignatures: false,
                                         });
--- a/js/packages/wallet-adapter-mobile/src/adapter.ts
+++ b/js/packages/wallet-adapter-mobile/src/adapter.ts
@@ -4,6 +4,7 @@ import {
     AuthorizationResult,
     AuthToken,
     Base64EncodedAddress,
+    Finality,
 } from '@solana-mobile/mobile-wallet-adapter-protocol';
 import {
     BaseMessageSignerWalletAdapter,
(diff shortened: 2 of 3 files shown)


js/packages/mobile-wallet-adapter-protocol-web3js/src/transact.ts
```diff
@@ -50,42 +50,20 @@ export async function transact<TReturn>(
...
+                                            transaction.feePayer = PublicKey.default;
                                         }
+                                        // Message serialization will fail unless there is a non-falsy blockhash.
                                         if (transaction.recentBlockhash == null) {
-                                            const { blockhash } = await getLatestBlockhashPromise();
-                                            transaction.recentBlockhash = blockhash;
+                                            transaction.recentBlockhash = '11111111111111111111111111111111';
                                         }
-                                        const serializedTransaction = transaction.serialize({
+                                        const serializedTransaction = await transaction.serialize({
```
steveluscher, line 66: Despite this, I have no idea how a wallet would know that these are garbage and should be replaced.

steveluscher: OK, I think either:

1. This was right before and I should close this PR, or
2. If someone sends a `Transaction` without a `feePayer` or `recentBlockhash` to `mobile-wallet-adapter-protocol-web3js` then we should _let_ it fail.

Thoughts?

jordansexton: `feePayer` seems straightforward to set to the expected account

Is the issue with setting `recentBlockhash` that we don't know what commitment level it should have?

js/packages/mobile-wallet-adapter-protocol-web3js/src/transact.ts
```diff
@@ -16,8 +16,8 @@ import { fromUint8Array, toUint8Array } from './base64Utils';
 
 interface Web3SignAndSendTransactionsAPI {
     signAndSendTransactions(params: {
-        connection: Connection;
-        fee_payer?: PublicKey;
+        commitment: Finality;
+        preflight_commitment: Finality;
```
jordansexton, line 20: We could presumably use whichever is the lower threshold of these to get the blockhash

jordansexton: The logic being, the blockhash may fail to be found if the `preflightCommitment` is `finalized` (the default) and the blockhash we're basing it on is only `confirmed` (seemingly the cause of https://github.com/brave/brave-browser/issues/24345)

js/packages/mobile-wallet-adapter-protocol-web3js/src/transact.ts
steveluscher: No, I just want to make `mobile-wallet-adapter-protocol-web3js` dumber. It does way too much. It's meant to be _only_ a wrapper around web3.js datatypes (ie. `mobile-wallet-adapter-protocol` takes a `payloads: Base64EncodedTransaction[]` while `mobile-wallet-adapter-protocol-web3js` takes a `transactions: Transaction[]`).

js/packages/mobile-wallet-adapter-protocol-web3js/src/transact.ts
steveluscher: The aim of this PR is to leave that work up to whoever calls into `mobile-wallet-adapter-protocol-web3js`.

steveluscher:
Updated this PR to choose the lesser of the two commitments.

js/packages/mobile-wallet-adapter-protocol-web3js/src/transact.ts
steveluscher: Which is to say I'll add this logic to the wallet adapter plugin.

steveluscher:
Stepping away; feel free to land this without me.

js/packages/mobile-wallet-adapter-protocol-web3js/src/transact.ts
jordansexton: Perhaps I'm missing something, but with the latest changes (which look good) moving things to WAM, are these args used anymore?

js/packages/wallet-adapter-mobile/src/adapter.ts
```diff
@@ -234,16 +235,51 @@ export class SolanaMobileWalletAdapter extends BaseMessageSignerWalletAdapter {
...
+                    let targetCommitment: Finality;
+                    switch (connection.commitment) {
+                        case 'confirmed':
+                        case 'finalized':
+                        case 'processed':
+                            targetCommitment = connection.commitment;
+                            break;
+                        default:
+                            targetCommitment = 'finalized';
+                    }
```
jordansexton, line 253:
```suggestion
                    const targetCommitment = connection.commitment || 'finalized';
```
This works, right?

js/packages/wallet-adapter-mobile/src/adapter.ts
```diff
@@ -234,16 +235,51 @@ export class SolanaMobileWalletAdapter extends BaseMessageSignerWalletAdapter {
...
+                        })(),
+                    ]);
+                    if (transaction.feePayer == null && this.publicKey) {
+                        transaction.feePayer = this.publicKey;
+                    }
                     const signatures = await wallet.signAndSendTransactions({
-                        connection,
-                        fee_payer: transaction.feePayer,
+                        commitment: targetCommitment,
+                        preflight_commitment: targetCommitment,
```
jordansexton, line 282: Don't think these are used by the callee anymore

js/packages/mobile-wallet-adapter-protocol-web3js/src/transact.ts
jordansexton: Ah, are these just passed through with `...rest`?

js/packages/wallet-adapter-mobile/src/adapter.ts
```diff
@@ -234,16 +235,51 @@ export class SolanaMobileWalletAdapter extends BaseMessageSignerWalletAdapter {
...
+                                        ? 1
+                                        : 0;
+                                const targetCommitmentScore =
+                                    targetCommitment === 'finalized' ? 2 : targetCommitment === 'confirmed' ? 1 : 0;
+                                const { blockhash } = await connection.getLatestBlockhash({
+                                    commitment:
+                                        preflightCommitment && preflightCommitmentScore < targetCommitmentScore
+                                            ? preflightCommitment
+                                            : targetCommitment,
+                                });
```
jordansexton, line 272: Assuming `preflightCommitment` and `targetCommitment` are the same value (reasonable default) or `preflightCommitment` isn't provided (also reasonable) could just check for truthy equality first and prefer `targetCommitment`

js/packages/wallet-adapter-mobile/src/adapter.ts
jordansexton, line 282: should this be `preflightCommitment`, or always use the same for both?

js/packages/wallet-adapter-mobile/src/adapter.ts
```diff
@@ -234,16 +235,51 @@ export class SolanaMobileWalletAdapter extends BaseMessageSignerWalletAdapter {
...
+                                transaction.recentBlockhash = blockhash;
+                            }
+                        })(),
+                    ]);
+                    if (transaction.feePayer == null && this.publicKey) {
+                        transaction.feePayer = this.publicKey;
+                    }
```
jordansexton, line 279:
```suggestion
                    transaction.feePayer = transaction.feePayer || this.publicKey;
```
It's going to fail anyway if neither feePayer or publicKey are set, so might as well get to it 😅

Style note -- do you prefer `||` or `??` if it doesn't matter (based on the types)? Is there a perf difference?

js/packages/wallet-adapter-mobile/src/adapter.ts
steveluscher: `commitment` is a `Commitment`. I wanted to narrow it to a `Finality`.

js/packages/wallet-adapter-mobile/src/adapter.ts
steveluscher: Is it not possible to have shenanigans like `preflightCommitment` being `processed` but `targetCommitment` being a higher commitment like `finalized`? That wouldn't work with the scheme above, right?

js/packages/wallet-adapter-mobile/src/adapter.ts
steveluscher: Damnit.

steveluscher: Fixed in [200b4bd](https://github.com/solana-mobile/mobile-wallet-adapter/commit/200b4bdccae240a9de8ea5486983654858c2fa5b).

js/packages/wallet-adapter-mobile/src/adapter.ts
steveluscher: Types is actually why I wrote it this way. `feePayer` doesn't accept `null` but `this.publicKey` might serve up `null`.

Here.

```typescript
transaction.feePayer ||= this.publicKey ?? undefined;
```

PR #233, opened 2022-09-09
use static string to test signMessage easier

nickcruz:
FYI it might be easier for other wallets / dapps to test signMessage if you have:
- a static / human-readable UTF-8 string (instead of random bytes that don't render nicely)
So instead of the UTF-8 encoding of the randomBytes from before (left), you get something like this (right)

Random bytes | static string
---|---
image|image

- a base58-representation of the signature bytes logged instead of the object stringified
So instead of this:
```
2022-09-09 11:41:35.551 14763-14812/com.solana.mobilewalletadapter.fakedapp D/MainViewModel: Signed message(s): SignPayloadsResult{signedPayloads=[[B@dd95339]}
```

you log this:
```
2022-09-09 11:41:35.555 14763-14812/com.solana.mobilewalletadapter.fakedapp D/MainViewModel: Signed message base58: 49NpST2nrCQtAcmVDAq7h1R5Tnp1aqyjzJKLCdKcBjQ9Z2mD2dr9B19dmHVJhTi5XDGCCBhBmxkeG4os7CTuwoyv
```


steveluscher (approved):
Seems reasonable! @sdlaver?

android/fakedapp/src/main/java/com/solana/mobilewalletadapter/fakedapp/MainViewModel.kt
```diff
@@ -432,7 +431,7 @@ class MainViewModel(application: Application) : AndroidViewModel(application) {
         var signedMessages: Array<ByteArray>? = null
         try {
             val result = client.signMessages(messages, addresses).get()
-            Log.d(TAG, "Signed message(s): $result")
+            Log.d(TAG, "Signed message base58: ${Base58EncodeUseCase(result.signedPayloads[0])}")
```
sdlaver, line 434: I'd love to print out all the payloads here. not just the 0'th one.
Perhaps leave the original Log statement (to dump the whole object, in case we add fields to it in the future), followed by iterating through each payload and printint out the base58 signature?

android/fakedapp/src/main/java/com/solana/mobilewalletadapter/fakedapp/MainViewModel.kt
```diff
@@ -181,10 +182,8 @@ class MainViewModel(application: Application) : AndroidViewModel(application) {
             if (!authorized) {
                 return@localAssociateAndExecute null
             }
-            val messages = Array(numMessages) {
-                Random.nextBytes(16384)
-            }
-            doSignMessages(client, messages, arrayOf(_uiState.value.publicKey!!))
+            val message = "To avoid digital dognappers, sign below to authenticate with CryptoCorgis.".toByteArray(Charsets.UTF_8)
+            doSignMessages(client, arrayOf(message), arrayOf(_uiState.value.publicKey!!))
```
sdlaver, line 186: This change will send only 1 message, irrespective of the value of the `numMessages` parameter. What if instead we did:
```
            val messages = Array(numMessages) { i ->
                "To avoid digital dognappers, sign below to authenticate with CryptoCorgis. [$i]".toByteArray(Charsets.UTF_8)
            }
            doSignMessages(client, messages, arrayOf(_uiState.value.publicKey!!))
```

This will append the message index to the string, ensuring that each message is unique (and thus that each signature will be unique as well)

sdlaver (changes requested):

sdlaver:
Superseded by #407

PR #243, opened 2022-09-20
Replace startActivityForResult with ActivityResultLauncher

ankur2136:
startActivityForResult is deprecated in favor of ActivityResultLauncher.

The sample isn't using the RequestCode. Future usage of ActivityResultsLaunched should include the resultCode inside the intent instead of passing it as a separate param.


sdlaver (approved):

steveluscher:
@ankur2136, can you take a second pass and find all the other callsites in this repository that need updating (eg. `SolanaMobileWalletAdapterModule.kt`).

PR #246, opened 2022-09-22
Refactor some pieces of AuthRepository.java

ankur2136:
1. Converted identityRecord to BuilderPattern. 

In this case, most fields are non-Nulls, however there would be other classes coming out of AuthRepository where it would make sense to use BuilderPattern. And I would like it to be consistent in the package. 

2. Created an abstract DbContentProvider class which can be extended by each Dao implementation. This class provides basic methods for query, insert, delete, rawQuery etc. 

3. Move a few methods for Identity table into an interface.

### DIFF
--- a/android/walletlib/src/main/java/com/solana/mobilewalletadapter/walletlib/authorization/AuthRepository.java
+++ b/android/walletlib/src/main/java/com/solana/mobilewalletadapter/walletlib/authorization/AuthRepository.java
@@ -177,33 +177,9 @@ public class AuthRepository {
         }
 
         // Look up the identity secret key for the key specified in this JWT
-        final IdentityRecord identityRecord;
-        try (final Cursor c = db.query(IdentityRecordSchema.TABLE_IDENTITIES,
-                IDENTITY_RECORD_COLUMNS,
-                IdentityRecordSchema.COLUMN_IDENTITIES_ID + "=?",
-                new String[] { identityIdStr },
-                null,
-                null,
-                null)) {
-            if (!c.moveToNext()) {
-                Log.w(TAG, "Identity not found: " + identityIdStr);
-                return null;
-            }
-
-            final int id = c.getInt(0);
-            final String name = c.getString(1);
-            final String uri = c.getString(2);
-            final String iconRelativeUri = c.getString(3);
-            final byte[] keyCiphertext = c.getBlob(4);
-            final byte[] keyIV = c.getBlob(5);
-            identityRecord = new IdentityRecord.IdentityRecordBuilder()
-                    .setId(id)
-                    .setName(name)
-                    .setUri(Uri.parse(uri))
-                    .setRelativeIconUri(Uri.parse(iconRelativeUri))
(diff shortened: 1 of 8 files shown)


oliveeyay (approved):

creativedrewy (approved):

android/walletlib/src/main/java/com/solana/mobilewalletadapter/walletlib/authorization/IdentityRecordDao.java
```diff
@@ -0,0 +1,97 @@
...
+            return new IdentityRecord.IdentityRecordBuilder()
+                    .setId(id)
+                    .setName(name)
+                    .setUri(Uri.parse(uri))
+                    .setRelativeIconUri(Uri.parse(iconRelativeUri))
+                    .setSecretKeyCiphertext(identityKeyCiphertext)
+                    .setSecretKeyIV(identityKeyIV)
+                    .build();
+        } catch (IllegalArgumentException e) {
+            return new IdentityRecord.IdentityRecordBuilder().build();
```
sdlaver, line 92: Hmm, silently swallowing an incorrect row is probably the wrong thing to do here. Perhaps allowing the `IllegalArgumentException` to propagate out is right thing to do, since this represents a programming error (i.e. an incorrectly configured query `Cursor` was provided)

android/walletlib/src/main/java/com/solana/mobilewalletadapter/walletlib/authorization/AuthRepository.java
```diff
@@ -359,45 +338,39 @@ public synchronized AuthRecord issue(@NonNull String name,
...
         }
 
         // If no matching identity exists, create one
         if (identityId == -1) {
             Log.d(TAG, "Creating IdentityRecord for " + name + '/' + uri + '/' + relativeIconUri);
 
             final Pair<byte[], byte[]> p = createEncryptedHmacSha256SecretKey();
-            identityKeyCiphertext = p.first;
-            identityKeyIV = p.second;
+            byte[] identityKeyCiphertext = p.first;
```
sdlaver, line 353: MINOR: these two variables can be final

android/walletlib/src/main/java/com/solana/mobilewalletadapter/walletlib/authorization/AuthRepository.java
```diff
@@ -99,6 +100,7 @@ private SQLiteDatabase ensureStarted() {
             }
 
             mAuthDb = new AuthDatabase(mContext, mAuthIssuerConfig);
+            mIdentityRecordDao = new IdentityRecordDao(mAuthDb.getWritableDatabase());
```
sdlaver, line 103: `getWritableDatabase` shouldn't be called from the application main thread: https://developer.android.com/reference/android/database/sqlite/SQLiteOpenHelper#getWritableDatabase()

`ensureDatabase` was used to lazily defer this to the caller thread (which was always done in an IO context, not Main context)

PR #254, opened 2022-09-24
feat: teach mobile-wallet-adapter to handle versioned transactions

steveluscher:
#### Problem

Versioned transactions will soon be enabled. Mobile Wallet Adapter is presently only equipped to serialize and deserialize legacy transactions.

#### Proposed solution

* Despite the fact that no wallets support versioned transactions yet, add the types and the implementation for serializing and deserializing both legacy and versioned transactions.
* Claim, at the `wallet-adapter-mobile` plugin level, that both types are supported, though it's impossible to know that without interrogating the actual wallet. This was done so that dApps don't have to upgrade later to be able to support versioned transactions.


js/packages/mobile-wallet-adapter-protocol-web3js/src/transact.ts
```diff
@@ -37,6 +44,27 @@ export interface Web3MobileWallet
...
+
+function getTransactionFromWireMessage(byteArray: Uint8Array): Transaction | VersionedTransaction {
+    const version = VersionedMessage.deserializeMessageVersion(byteArray);
+    if (version === 'legacy') {
+        return Transaction.from(byteArray);
+    } else {
+        return VersionedTransaction.deserialize(byteArray);
+    }
+}
+
```
steveluscher, line 67: @jstarry, does this look right to you? See below where these are used, to serialize/deserialize transactions to/from mobile wallets.

js/packages/wallet-adapter-mobile/src/adapter.ts
```diff
@@ -242,51 +256,62 @@ export class SolanaMobileWalletAdapter extends BaseMessageSignerWalletAdapter {
...
-                                transaction.recentBlockhash = blockhash;
-                            }
-                        })(),
+                        'version' in transaction
+                            ? null
+                            : /**
+                               * Unlike versioned transactions, legacy `Transaction` objects
+                               * may not have an associated `feePayer` or `recentBlockhash`.
+                               * This code exists to patch them up in case they are missing.
+                               */
```
steveluscher, line 267: cc/ @jstarry, @jordansexton. Correct?

js/packages/wallet-adapter-mobile/src/adapter.ts
```diff
@@ -44,6 +52,10 @@ function getPublicKeyFromAddress(address: Base64EncodedAddress): PublicKey {
 }
 
 export class SolanaMobileWalletAdapter extends BaseMessageSignerWalletAdapter {
+    readonly supportedTransactionVersions: Set<TransactionVersion> = new Set(
+        // FIXME(#244): We can't actually know what versions are supported until we know which wallet we're talking to.
+        ['legacy', 0],
+    );
```
steveluscher, line 58: cc/ @jstarry, @sdlaver. This is the awkward bit; MWA plugin doesn't know what transaction versions it supports until it actually associates with a particular mobile wallet app. Despite that, I thought it impossible to do anything other than to tell the _type system_ that we support all versions.

steveluscher:
If there's no feedback on this in the next little bit, I'm going to ship it, mostly because I need it for the next step which is to bake mobile wallet adapter into `@solana/wallet-adapter`. Also, because package managers hate library developers and `pnpm link` / `yarn link` just don't play well together.

Issue #94, opened 2022-07-11
If you trigger an association activity while the last one is still concluding, it will silently fail


steveluscher:
#### Problem

Naturally, the correct thing to do when you have multiple operations to perform with a wallet is to:

1. Open a session through association
2. Invoke _n_ API methods
3. Close the session

Despite this, we know that people – for one reason or another – are going to write apps that execute that _full_ loop multiple times, with one call coming on the heels of the next.

I discovered a bug with doing so when testing the React Native SDK. I called two `@solana/wallet-adapter` methods in a row and observed the second one being silently dropped. This appears to be because, if we start an activity before the first one has _fully_ animated back to the original app, that activity will be ignored.

#### Repro

Make the following modification:

```diff
diff --git a/android/fakedapp/src/main/java/com/solana/mobilewalletadapter/fakedapp/MainViewModel.kt b/android/fakedapp/src/main/java/com/solana/mobilewalletadapter/fakedapp/MainViewModel.kt
index 8ffb5a6..1e286f5 100644
--- a/android/fakedapp/src/main/java/com/solana/mobilewalletadapter/fakedapp/MainViewModel.kt
+++ b/android/fakedapp/src/main/java/com/solana/mobilewalletadapter/fakedapp/MainViewModel.kt
@@ -78,9 +78,12 @@ class MainViewModel(application: Application) : AndroidViewModel(application) {
     }
 
     suspend fun authorizeAndSignTransaction(sender: StartActivityForResultSender) {
+        var authorized = localAssociateAndExecute(sender) { client ->
+            doAuthorize(client)
+        }
+        // delay(1000) <--- add this to give time for the activity to finish, and it works.
         val signedTransactions = localAssociateAndExecute(sender) { client ->
-            val authorized = doAuthorize(client)
-            if (authorized) {
+            if (authorized == true) {
                 val transactions = Array(1) {
                     MemoTransaction.create(uiState.value.publicKeyBase58!!)
                 }
```

Open the fake dapp and click the ‘combined authorize and sign’ button.

sdlaver:
Root cause: you can't start activities while not in the foreground. Back to back calls must wait for the originating activity to return to the `RESUMED` state. This should be elevated to an immediate exception, to catch the programming error.

steveluscher:
What if, instead, `localAssociateAndExecute` just didn't resume until the app returns to the foreground?

steveluscher:
[...]

PR #264, opened 2022-09-27
Wait for the activity to be fully resume before returning the result

ankur2136:
#94 If you trigger an association activity while the last one is still concluding, it will silently fail

Inserting a wait at the end of the `localAssociateAndExecute` method. This lock basically waits till the ActivityResultCallback is not received. That guarantees that the original activity has resumed before using the results from the previous activity

### DIFF
--- a/android/fakedapp/src/main/java/com/solana/mobilewalletadapter/fakedapp/MainViewModel.kt
+++ b/android/fakedapp/src/main/java/com/solana/mobilewalletadapter/fakedapp/MainViewModel.kt
@@ -32,6 +32,8 @@ import java.util.concurrent.CancellationException
 import java.util.concurrent.ExecutionException
 import java.util.concurrent.TimeUnit
 import java.util.concurrent.TimeoutException
+import java.util.concurrent.locks.ReentrantLock
+import kotlin.concurrent.withLock
 import kotlin.random.Random
 
 class MainViewModel(application: Application) : AndroidViewModel(application) {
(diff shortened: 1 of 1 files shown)


android/fakedapp/src/main/java/com/solana/mobilewalletadapter/fakedapp/MainViewModel.kt
```diff
@@ -40,6 +42,9 @@ class MainViewModel(application: Application) : AndroidViewModel(application) {
 
     private val mobileWalletAdapterClientSem = Semaphore(1) // allow only a single MWA connection at a time
 
+    private val lock = ReentrantLock()
```
sdlaver, line 45: Do we need the reentrancy aspect of this lock? A Kotlin coroutines [mutex](https://kotlinlang.org/api/kotlinx.coroutines/kotlinx-coroutines-core/kotlinx.coroutines.sync/-mutex/) is also an option here, that integrates better with coroutines.

android/fakedapp/src/main/java/com/solana/mobilewalletadapter/fakedapp/MainViewModel.kt
```diff
@@ -519,6 +538,7 @@ class MainViewModel(application: Application) : AndroidViewModel(application) {
             try {
                 sender.startActivityForResult(associationIntent) {
                     viewModelScope.launch {
+                        releaseLock()
```
sdlaver, line 541: We'll need to rework the callback structure a bit to move the invocation of this closure into ViewModel, to avoid any lifecycle issues stemming from transient Activity re-creation.

sdlaver (changes requested):

PR #260, opened 2022-09-27
Update incoming sequence number handling to match latest spec revision

sdlaver:
Fixes #257, fixes #258, fixes #259

PR #266, opened 2022-09-28
Use strict inequality operator for sequence number check

sdlaver:


ankur2136 (approved):

steveluscher:
There was another file in the last PR, yeah?

steveluscher (changes requested):
There was another file in the last PR, yeah?

sdlaver:
> There was another file in the last PR, yeah?

https://github.com/solana-mobile/mobile-wallet-adapter/pull/260
There was a Java file as well. Java's two equality tests are:
- [`Object.equals`](https://docs.oracle.com/javase/7/docs/api/java/lang/Object.html#equals(java.lang.Object))
- `==`

The types being compared are `int` and `int`, hence `==` is the appropriate comparison.

steveluscher (approved):
Guh. Sorry. Both times I was reading your PRs I was on mobile, where positively every useful bit of context is cut off in the UI in some way.

PR #267, opened 2022-09-28
spec: Wallets can report the transaction versions they support

### LINKED ISSUE #244: Add a `supported_transaction_versions` field to the `get_capabilities` response
With the introduction of txv2, MWA implementations should enumerate what versions of Solana transactions they support

sdlaver:
Fixes #244

### DIFF
--- a/spec/spec.md
+++ b/spec/spec.md
@@ -503,6 +503,7 @@ get_capabilities
     "supports_sign_and_send_transactions": <supports_sign_and_send_transactions>,
     "max_transactions_per_request": <max_transactions_per_request>,
     "max_messages_per_request": <max_messages_per_request>,
+    "supported_transaction_versions": [<supported_transaction_versions>, ...]
 }
 ```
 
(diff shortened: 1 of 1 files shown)


sdlaver:
Before merging this PR, I will file issues for each of the implementations that need to be updated to support this field.

steveluscher (approved):
🤞🏻

sdlaver:
Filed issues #270, #271

jordansexton (approved):
Looks good! Should the field be optional or required? If optional, not being provided could default to mean only `legacy` transactions are supported.

spec/spec.md
```diff
@@ -503,6 +503,7 @@ get_capabilities
     "supports_sign_and_send_transactions": <supports_sign_and_send_transactions>,
     "max_transactions_per_request": <max_transactions_per_request>,
     "max_messages_per_request": <max_messages_per_request>,
+    "supported_transaction_versions": [<supported_transaction_versions>, ...]
```
jordansexton, line 506: nit / not sure we should do anything about this, but the other fields use `supports_` and this one uses `supported_`

sdlaver:
> Should the field be optional or required?

My thinking was that this should be required - knowing this is important for dapps, and requiring it in the protocol allows me to require it in the `walletlib` implementation, forcing wallets to think about this. If I support a default, it might be missed, and that would lead to the suboptimal case of a dapp wanting to use v0 transactions with a wallet but reporting an error.

That said, in the client implementations, I plan to treat this as optional for the time being (and assume `legacy` if not present). This will provide time for wallets to upgrade `walletlib` to a version supporting this field.

jordansexton:
> My thinking was that this should be required

Seems fine. FWIW I remembered that I also opted to require it in the standard interface: https://github.com/wallet-standard/wallet-standard/blob/078f8d56b060de3f248cd76a6cb592100cf58415/packages/solana/features/src/solanaSignTransaction.ts#L12

PR #273, opened 2022-09-29
Fix some misc minor warnings

sdlaver:


android/clientlib-ktx/src/main/java/com/solana/mobilewalletadapter/clientlib/MobileWalletAdapter.kt
```diff
@@ -81,14 +89,6 @@ class MobileWalletAdapter(
                     Log.e(TAG, "JSON-RPC client exception", cause)
                     throw e
                 }
-                is MobileWalletAdapterClient.InvalidPayloadsException -> {
```
sdlaver, line 84: These were after an `is JsonRpc20Client.JsonRpc20RemoteException` case, which is a superclass and would capture all instances of these two exceptions.

ankur2136 (approved):

PR #291, opened 2022-10-14
Add a deauthorize callback, to inform wallets when a dapp is deauthorized

### LINKED ISSUE #290: `walletlib` should include a callback for `deauthorize`

sdlaver:
Fixes #290

### DIFF
--- a/android/walletlib/src/main/java/com/solana/mobilewalletadapter/walletlib/scenario/Scenario.java
+++ b/android/walletlib/src/main/java/com/solana/mobilewalletadapter/walletlib/scenario/Scenario.java
@@ -255,7 +255,17 @@ public abstract class Scenario {
                     mActiveAuthorization = null;
                 }
             }
-            mIoHandler.post(() -> request.complete(null));
+
+            if (authRecord != null) {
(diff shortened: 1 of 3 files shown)


android/walletlib/src/main/java/com/solana/mobilewalletadapter/walletlib/scenario/Scenario.java
```diff
@@ -332,6 +342,7 @@ public interface Callbacks {
         // Request callbacks
         void onAuthorizeRequest(@NonNull AuthorizeRequest request);
         void onReauthorizeRequest(@NonNull ReauthorizeRequest request);
+        void onDeauthorizeRequest(@NonNull DeauthorizeRequest request);
```
Sinishisimus, line 345: Maybe we should name the callback onDeauthorizedRequest to reflect the fact that the user has already been deauthorised?

sdlaver: 👍 will rename

Sinishisimus (approved):
Looks good.

PR #297, opened 2022-10-17
Update the README of the `wallet-adapter` plugin to make it clear how this should interact with `@solana/wallet-adapter-react`

steveluscher:


steveluscher:
Does this make sense, @jordansexton?

jordansexton:
Yeah, looks good. We can link this from WA

PR #302, opened 2022-10-18
Implement `getCapabilites` in JS protocol

### LINKED ISSUE #271: Implement `supported_transaction_versions` in JS

steveluscher:
Closes #271


js/packages/mobile-wallet-adapter-protocol/src/types.ts
```diff
@@ -63,6 +65,18 @@ export interface CloneAuthorizationAPI {
...
 }
+
+export interface GetCapabilitiesAPI {
+    getCapabilities(): Promise<
+        Readonly<{
+            supports_clone_authorization: boolean;
+            supports_sign_and_send_transactions: boolean;
+            max_transactions_per_request: boolean;
+            max_messages_per_request: boolean;
+            supported_transaction_versions: ReadonlyArray<TransactionVersion>;
```
steveluscher, line 76: @jordansexton, anything you see particularly bad about this being a `ReadonlyArray` instead of a `Set`? I wanted to avoid this PR being anything other than `TypeScript`.

jordansexton: We use ReadonlyArray in the standard too, I don't really care for Set either. I assume whatever uniqueness/perf benefit it might have is likely offset for small sets.

jordansexton: In terms of implementation, I also prefer arrays since we want to enforce them being readonly, and cloning sets doesn't have a native API so we just end up spreading them to copy, which is probably bad for perf.

PR #303, opened 2022-10-18
Shim sign and send for those that do not support it

steveluscher:
If a wallet does not support `signAndSendTransactions` then the mobile wallet adapter plugin will now use the `signTransactions` method to obtain a signed transaction, then send that transaction on the client side.

#### Test plan

1. Recompile `fakewallet` setting `supportsSignAndSendTransactions` to `false`
2. Use the example web app to send a transaction

https://user-images.githubusercontent.com/13243/196574787-a7f15358-c98b-4c92-aa13-a481401f3e08.mp4

### DIFF
--- a/js/packages/wallet-adapter-mobile/src/adapter.ts
+++ b/js/packages/wallet-adapter-mobile/src/adapter.ts
@@ -53,6 +53,12 @@ function getPublicKeyFromAddress(address: Base64EncodedAddress): PublicKey {
     return new PublicKey(publicKeyByteArray);
 }
 
+function isVersionedTransaction(
+    transaction: LegacyTransaction | VersionedTransaction,
+): transaction is VersionedTransaction {
+    return 'version' in transaction;
+}
+
 export class SolanaMobileWalletAdapter extends BaseMessageSignerWalletAdapter {
     readonly supportedTransactionVersions: Set<TransactionVersion> = new Set(
         // FIXME(#244): We can't actually know what versions are supported until we know which wallet we're talking to.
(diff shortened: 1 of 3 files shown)


steveluscher:
This is not yet tested, because I haven't installed such a wallet. I'll have to hack up `fakewallet` to disable its transaction sending capabilities.

js/packages/wallet-adapter-mobile/src/adapter.ts
```diff
@@ -304,7 +304,43 @@ export class SolanaMobileWalletAdapter extends BaseMessageSignerWalletAdapter {
...
+                                ? 1
+                                : 0;
+                        const targetCommitmentScore =
+                            targetCommitment === 'finalized' ? 2 : targetCommitment === 'confirmed' ? 1 : 0;
+                        return preflightCommitmentScore < targetCommitmentScore
+                            ? targetPreflightCommitment
+                            : targetCommitment;
+                    }
+                    const [capabilities, _1, _2] = await Promise.all([
+                        wallet.getCapabilities(),
```
steveluscher, line 343: Presumably there's nothing that should prohibit a client from firing off `get_capabilities` and `reauthorize` at the same time, right @sdlaver?

js/packages/wallet-adapter-mobile/src/adapter.ts
```diff
@@ -316,55 +352,31 @@ export class SolanaMobileWalletAdapter extends BaseMessageSignerWalletAdapter {
...
-                                              ? 2
-                                              : targetCommitment === 'confirmed'
-                                              ? 1
-                                              : 0;
                                       const { blockhash } = await connection.getLatestBlockhash({
-                                          commitment:
-                                              preflightCommitmentScore < targetCommitmentScore
-                                                  ? targetPreflightCommitment
-                                                  : targetCommitment,
+                                          commitment: getTargetCommitment(),
```
jordansexton, line 356: Probably should pass minContextSlot to getLatestBlockhash also. But if there is no minContextSlot to pass, we should arguably call getLatestBlockhashAndContext instead and start using the context slot we get back as the min.

js/packages/wallet-adapter-mobile/src/adapter.ts
```diff
@@ -316,55 +352,31 @@ export class SolanaMobileWalletAdapter extends BaseMessageSignerWalletAdapter {
...
+                            minContextSlot,
+                            transactions: [transaction],
+                        });
+                        return signatures[0];
+                    } else {
+                        const signedTransaction = await this.signTransaction(transaction);
+                        if ('version' in signedTransaction) {
+                            return await connection.sendTransaction(signedTransaction);
+                        } else {
+                            const serializedTransaction = (signedTransaction as LegacyTransaction).serialize();
```
jordansexton, line 373: Can use a type guard for `is VersionedTransaction`

js/packages/wallet-adapter-mobile/src/adapter.ts
```diff
@@ -304,7 +304,43 @@ export class SolanaMobileWalletAdapter extends BaseMessageSignerWalletAdapter {
             const minContextSlot = options?.minContextSlot;
             try {
                 return await this.transact(async (wallet) => {
-                    await Promise.all([
+                    function getTargetCommitment() {
```
jordansexton, line 307: We should probably add this logic to Wallet Adapter's prepareTransaction method...

js/packages/wallet-adapter-mobile/src/adapter.ts
steveluscher: Sweet. Will do this in a follow up PR.

js/packages/wallet-adapter-mobile/src/adapter.ts
steveluscher: 😬

PR #304, opened 2022-10-19
Ignore outcome of irrelevant transacts

steveluscher:
Once you've made the move to disconnect, any in-flight `transact` calls are no longer relevant. Silently ignore the return values and any exceptions thrown from them.

#### Test plan

1. Move to connect
2. Without interacting with the wallet, switch back to the web browser
3. Disconnect

Observe that you can click to connect again, and that timeout errors from the previous connection attempt _do not surface_.

https://user-images.githubusercontent.com/13243/196566996-2b719a20-7ae1-4fb6-9fcd-113940467465.mov

Addresses https://github.com/solana-labs/wallet-adapter/issues/611

### DIFF
--- a/js/packages/wallet-adapter-mobile/src/adapter.ts
+++ b/js/packages/wallet-adapter-mobile/src/adapter.ts
@@ -68,6 +68,12 @@ export class SolanaMobileWalletAdapter extends BaseMessageSignerWalletAdapter {
     private _authorizationResult: AuthorizationResult | undefined;
     private _authorizationResultCache: AuthorizationResultCache;
     private _connecting = false;
+    /**
+     * Every time the connection is recycled in some way (eg. `disconnect()` is called)
+     * increment this and use it to make sure that `transact` calls from the previous
+     * 'generation' don't continue to do work and throw exceptions.
+     */
+    private _connectionGeneration: number = 0;
     private _cluster: Cluster;
     private _onWalletNotFound: (mobileWalletAdapter: SolanaMobileWalletAdapter) => Promise<void>;
     private _publicKey: PublicKey | undefined;
@@ -242,6 +248,8 @@ export class SolanaMobileWalletAdapter extends BaseMessageSignerWalletAdapter {
 
     async disconnect(): Promise<void> {
         this._authorizationResultCache.clear(); // TODO: Evaluate whether there's any threat to not `awaiting` this expression
+        this._connecting = false;
+        this._connectionGeneration++;
         delete this._authorizationResult;
         delete this._publicKey;
         delete this._selectedAddress;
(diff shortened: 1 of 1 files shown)


jordansexton:
I'm not sure if this is possible, but what about making it so that if the context switches back to the app without completing the connect step on the wallet side, it rejects the pending promise in the app? To me this is akin to closing the wallet popup.

I'm thinking about this from the perspective of a

jordansexton:
I'm thinking about this from the perspective of a dapp, but it's probably not going to have a separate "disconnect" button like we do in the example. Most dapps already don't, and especially with screen real estate limited on mobile, I see very low chance of it.

steveluscher:
> …what about making it so that if the context switches back to the app without completing the connect step on the wallet side, it rejects the pending promise in the app?

100%. Wallets are _supposed_ to do this by closing the WebSocket connection when they switch back, but I've observed some _not_ doing that. This is a defense against that.

Also, a user could get the Android disambiguation dialog, and just close it. Nothing we can learn from that; it'll just time out. I want to give the user at least the opportunity to press disconnect and try again, without the ghost of the timeout coming back to haunt them.

jordansexton:
> a user could get the Android disambiguation dialog, and just close it. Nothing we can learn from that

Ah, we don't get something like window blur or focus events? I have to apologize, I'm not very aware of the tools at our disposal.

steveluscher:
> Ah, we don't get something like window blur or focus events?

Mmmm… maybe actually. The thing I have to be careful about is getting into a race with a wallet session _actually concluding successfully_ – I don't want a focus event on the window to ‘beat’ a session closing successfully, lest I accidentally treat a successful session as having failed.

js/packages/wallet-adapter-mobile/src/adapter.ts
```diff
@@ -251,9 +259,13 @@ export class SolanaMobileWalletAdapter extends BaseMessageSignerWalletAdapter {
     private async transact<TReturn>(callback: (wallet: Web3MobileWallet) => TReturn): Promise<TReturn> {
         const walletUriBase = this._authorizationResult?.wallet_uri_base;
         const config = walletUriBase ? { baseUri: walletUriBase } : undefined;
+        const currentConnectionGeneration = this._connectionGeneration;
         try {
             return await transact(callback, config);
         } catch (e) {
+            if (this._connectionGeneration !== currentConnectionGeneration) {
+                await new Promise(() => {}); // Never resolve.
```
jordansexton, line 267: I've never seen this kind of pattern (infinitely hanging promise) so I'm not sure what this does in this context. If a promise never resolves, does stuff hang around without getting GC'd, or stuck in any kind of weird state?

PR #305, opened 2022-10-19
Match the slot when fetching the blockhash and simulating the transaction

steveluscher:
#### Problem

* If a min slot was specified but a blockhash was _not_ we have been fetching the blockhash at any old slot.
* If neither a min slot nor a blockhash were specified, we could use to fetch the min slot along with the blockhash, and forward that along to `sendTransaction`

#### Summary of changes

* ✅ slot ❌ blockhash: specify the min-slot when fetching the blockhash
* ❌ slot ❌ blockhash: fetch the slot along with the blockhash and use it when sending the transaction

### DIFF
--- a/js/packages/wallet-adapter-mobile/src/adapter.ts
+++ b/js/packages/wallet-adapter-mobile/src/adapter.ts
@@ -22,6 +22,7 @@ import {
     WalletSignTransactionError,
 } from '@solana/wallet-adapter-base';
 import {
+    Blockhash,
     Connection,
     PublicKey,
     SendOptions,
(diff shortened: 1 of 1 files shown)


steveluscher:
Note that this will make this API _throw errors_ when the min slot isn't met, which I guess is more correct, but probably also more annoying. Thoughts on this tradeoff?

steveluscher:
> I'm not actually sure what the semantics of this feature mean in the RPC.

The RPC just fatals without doing any work if it hasn't already moved past that slot.

https://github.com/solana-labs/solana/blob/5d172151a2d3d652c6a2e089b09e6fca7f2aa50e/rpc/src/rpc.rs#L1592-L1597

steveluscher:
@jstarry, what are your thoughts on this?

* Before: you could simulate a transaction against the ‘wrong’ slot, but it could still succeed
* After: the operation will fail for sure, not because the transaction would have failed, but just by virtue of the wrongness of the slot

I feel like this change, while ‘correct,’ could reduce perceived reliability overall. Maybe we should honor the slot restriction _iff_ the user has supplied a `minContextSlot`, but otherwise just cross our fingers and hope?

PR #329, opened 2022-11-17
feat: expose origin of connection

kewde:
This PR exposes the `Origin` of a connection in the `LocalWebSocketServerScenario` class.
This capability allows wallets to enforce a more rigorous security policy, enabling them to enforce a constraint that if an origin is provided, it must match its claimed identity. This is in no way a replacement for a proper web dApp attestation schema, merely allowing us to create another security hurdle for malicious web dApps that aim to imitate legitimate web dApps.
This PR has not been tested yet.

### DIFF
--- a/android/walletlib/src/main/java/com/solana/mobilewalletadapter/walletlib/scenario/LocalWebSocketServerScenario.java
+++ b/android/walletlib/src/main/java/com/solana/mobilewalletadapter/walletlib/scenario/LocalWebSocketServerScenario.java
@@ -7,6 +7,7 @@ package com.solana.mobilewalletadapter.walletlib.scenario;
 import android.content.Context;
 
 import androidx.annotation.NonNull;
+import androidx.annotation.Nullable;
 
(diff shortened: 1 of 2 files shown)


sdlaver:
I like the direction of this, but I'd request a change in how the wallet access this data. To keep the wallet simple, I propose that the origin field be conveyed to it via the `AuthorizeRequest` or `ReauthorizeRequest` classes. Something like this:

1. Define an interface, `OriginRequest`, with a getter
2. Implement the `OriginRequest` interface on `AuthorizeRequest` and `ReauthorizeRequest`, backed by a class member
3. Capture the origin (via an empty method?) in `Scenario.onSessionEstablished`, and clear it in `Scenario.onSessionClosed`
4. Override the capture method in `LocalWebSocketServerScenario` to capture the origin from the websocket server

This will keep the data flow of the origin parameter aligned with the authorize/reauthorize request, and make it easy to consume.

WDYT?

kewde:
@Funkatronics No, sadly not. The plan forward however has been laid out. I will close this PR, feel free to open an issue with the plan above.

PR #334, opened 2022-11-29
docs: add warning about power saving mode

kewde:
This section adds a warning about the power saving mode and its implications.


Funkatronics:
going to merge this then add more info about the low power no connection callback in a separate PR

Issue #344, opened 2022-12-14
`fakedapp` should be able to create versioned transactions (for testing purposes)


sdlaver:
1. Add a dropdown to `fakedapp` to allow selecting transaction format to create (legacy or v0)
2. Rename existing `MemoTransactionuseCase` class to `MemoTransactionLegacyUseCase`
3. Extract an interface from `MemoTransactionLegacyUseCase`
4. Implement `MemoTransactionV0UseCase`
5. Validate with existing MWA wallets (e.g. Phantom, Solflare) that support V0 transactions (see the output of the `get_capabilities` MWA request to verify) and with `fakewallet` on Saga and on Android simulator (to check both in-wallet signing and Seed Vault signing)

Funkatronics:
I have the V0 message format working, tested so far with both phantom and the fake wallet (with some added support for signing V0 txns). Just need to button everything up and attach stuff to the UI.

Funkatronics:
fixed in #346

PR #346, opened 2022-12-21
`fakedapp` Versioned Transactions Support

Funkatronics:
## Task
- Add support for versioned transactions in the `fakedapp`
- https://github.com/solana-mobile/mobile-wallet-adapter/issues/344

## Work Completed
- created abstract `MemoTransactionUseCase`  with common memo transaction code
- created both `MemoTransactionLegacyUseCase`  and  `MemoTransactionV0UseCase`  with relevant transaction templates
- added "Transaction Version" drop down spinner to the activity and connected everything in the vm

## `fakewallet` changes
- added support for versioned transactions in the fake wallet (checks transaction message bit flag and reads version) 

### `fakewallet`
- verified signing both legacy and v0 transactions with the fake wallet completes without error
- verified signing and sending both legacy and v0 transactions to testnet ([legacy](https://solscan.io/tx/5pCZzohoz7RVKnnPEskf5jj7HXRHWaWNDBDqyzn19RNZ6DVDfXHjxtqJUDsX62fdWNf98o8V2sSFChXGxhkkuxEf?cluster=testnet) and [v0](https://solscan.io/tx/2K8W3zeWtEiWonDRNbyvxUbpZGetGDV2TL6vYoQTU1LZEfD5FVBL87PMEuhK5gxxReVcmJ8edn9gK3GP2HWRnqcq?cluster=testnet))

### Phantom
- with seed vault:
  - verified signing both legacy and v0 transactions with Phantom completes without error
  - verified signing and sending both legacy and v0 transactions to testnet
- with internal Phantom wallet
  - verified signing both legacy and v0 transactions with Phantom completes without error
  - ~signing and sending both legacy and v0 transactions to testnet~ FAIL: Phantom returns error "User did not authorize signing"

### Solflare
- with seed vault:
  - verified signing both legacy and v0 transactions with Solflare completes without error
  - cannot sign and send transactions: Solflare returns error "method 'sign_and_send_transactions' not available"

### DIFF
--- a/android/fakedapp/src/main/java/com/solana/mobilewalletadapter/fakedapp/MainViewModel.kt
+++ b/android/fakedapp/src/main/java/com/solana/mobilewalletadapter/fakedapp/MainViewModel.kt
@@ -37,11 +37,10 @@ class MainViewModel(application: Application) : AndroidViewModel(application) {
     private val _uiState = MutableStateFlow(UiState())
     val uiState = _uiState.asStateFlow()
 
-    // should probably make an enum or similar abstraction but this is fine for now
-    val supportedTxnVersions = listOf("legacy", "v0")
+    val supportedTxnVersions = listOf(MemoTransactionVersion.Legacy, MemoTransactionVersion.V0)
     val transactionUseCase get() = when(_uiState.value.txnVersion) {
-        "v0" -> MemoTransactionV0UseCase
-        else -> MemoTransactionLegacyUseCase
+        MemoTransactionVersion.Legacy -> MemoTransactionLegacyUseCase
+        MemoTransactionVersion.V0 -> MemoTransactionV0UseCase
     }
 
     private val mobileWalletAdapterClientSem = Semaphore(1) // allow only a single MWA connection at a time
@@ -101,7 +100,7 @@ class MainViewModel(application: Application) : AndroidViewModel(application) {
         }
     }
 
-    fun setTransactionVersion(txnVersion: String) {
+    fun setTransactionVersion(txnVersion: MemoTransactionVersion) {
         _uiState.update { it.copy(txnVersion = txnVersion) }
     }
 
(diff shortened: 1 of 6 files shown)


android/fakedapp/src/main/res/values/strings.xml
```diff
@@ -24,6 +24,9 @@
     <string name="label_sign_msg_x1">Sign msg x1</string>
     <string name="label_sign_msg_x3">x3</string>
     <string name="label_sign_msg_x20">x20</string>
+    <string name="label_txn_version">Transaction Version:</string>
+    <string name="label_txn_version_legacy">Legacy</string>
```
sdlaver, line 28: NIT: in the convention used in this test app (which is admittedly entirely arbitrary), `label_` is a prefix for widget labels, and `string_` is ususally used for, e.g., the contents of widgets.

android/fakedapp/src/main/java/com/solana/mobilewalletadapter/fakedapp/MainViewModel.kt
```diff
@@ -39,6 +37,13 @@ class MainViewModel(application: Application) : AndroidViewModel(application) {
     private val _uiState = MutableStateFlow(UiState())
     val uiState = _uiState.asStateFlow()
 
+    // should probably make an enum or similar abstraction but this is fine for now
```
sdlaver, line 40: Since `fakedapp` is our most canonical example of how to use `clientlib`, I'd recommend we make an enum for this, and do the mapping in `MainActivity`. There's a decent chance that much of this sample gets copy-pasted elsewhere, so worth the extra verbosity to do it cleanly.

sdlaver (approved):

android/fakedapp/src/main/java/com/solana/mobilewalletadapter/fakedapp/MainViewModel.kt
Funkatronics: makes sense, will do!

sdlaver (approved):

PR #352, opened 2023-01-11
fix: allow re-using port

kewde:
This should allow force binding to a port that has connections in timeout. (Untested)


sdlaver:
@kewde can I ask, how were you able to repro this issue?

Per the [spec](https://solana-mobile.github.io/mobile-wallet-adapter/spec/spec.html#local-uri) and the [clientlib implementation](https://github.com/solana-mobile/mobile-wallet-adapter/blob/main/android/clientlib/src/main/java/com/solana/mobilewalletadapter/clientlib/scenario/LocalAssociationScenario.java#L61), port numbers should be randomized on each connection. Collisions should be exceeding rare (random selection from 16384 ports in the dynamic port range).

kewde:
@sdlaver 

Closing because it was using stale data, Turned out to be a bug with `ReactActivity`. The activity held the original intent and thus URL. Listening for new intents and prioritizing those resolved the issue.

PR #355, opened 2023-01-12
publish clientlib-ktx to maven

ankur2136:


creativedrewy (approved):

android/clientlib-ktx/build.gradle
```diff
@@ -33,6 +35,84 @@ android {
...
+    }
+}
+
+publishing {
+    publications {
+        release(MavenPublication) {
+            groupId = group
+            artifactId = 'mobile-wallet-adapter-clientlib-ktx'
+            pom {
+                name = 'Mobile Wallet Adapter - Client library'
```
sdlaver, line 53: We should probably update the `name` and `description` fields to be different than `clientlib`

android/clientlib-ktx/build.gradle
```diff
@@ -33,6 +35,84 @@ android {
...
+            withJavadocJar()
+        }
+    }
+}
+
+publishing {
+    publications {
+        release(MavenPublication) {
+            groupId = group
+            artifactId = 'mobile-wallet-adapter-clientlib-ktx'
```
sdlaver, line 51: Along with `groupId = group` and `artifactId = 'mobile-wallet-adapter-clientlib-ktx'`, we might want to add `version = version + '-alpha'`, unless we are sufficiently satisfied with the API surface that we are willing to commit to it. @creativedrewy thoughts?

sdlaver (changes requested):

android/clientlib-ktx/build.gradle
ankur2136: Not needed since this is already used in a couple of sample apps although not as a dependency

android/clientlib-ktx/build.gradle
ankur2136: Done.

sdlaver (approved):

PR #362, opened 2023-01-20
Fix wallet select disambiguation cancellation crash

creativedrewy:
Clientlib KTX unfortunately has had a bug all along where if you cancel the wallet selection disambiguation dialog, after the full socket connection timeout, the lib would crash.

We had initially implemented KTX MWA to propagate any exceptions to the conusmer, however this was a design flaw. As this is a convenience library, any underlying exceptions should not crash the consumer application.

The new design implements a returned sealed class hierarchy where if the transact operation was successful, the consumer gets a `Success`  object with a data payload. If there was an underlying failure, the consumer gets a `Failure` object with relevant error details. The consumer can decide if they want to parse/respond to certain errors. This sealed class hierarchy could be extended to have specific error types that we want to notify the user about.

### DIFF
--- a/android/clientlib-ktx/src/main/java/com/solana/mobilewalletadapter/clientlib/ActivityResultSender.kt
+++ b/android/clientlib-ktx/src/main/java/com/solana/mobilewalletadapter/clientlib/ActivityResultSender.kt
@@ -1,7 +1,35 @@
 package com.solana.mobilewalletadapter.clientlib
 
 import android.content.Intent
+import androidx.activity.ComponentActivity
+import androidx.activity.result.ActivityResultLauncher
+import androidx.activity.result.contract.ActivityResultContracts
+import androidx.annotation.GuardedBy
 
-interface ActivityResultSender {
-    fun launch(intent: Intent)
+class ActivityResultSender(
+    rootActivity: ComponentActivity
+) {
+    @GuardedBy("this")
+    private var callback: (() -> Unit)? = null
+
+    private val activityResultLauncher: ActivityResultLauncher<Intent> =
+        rootActivity.registerForActivityResult(ActivityResultContracts.StartActivityForResult()) {
+            onActivityComplete()
+        }
+
+    fun startActivityForResult(intent: Intent, onActivityCompleteCallback: () -> Unit) {
+        synchronized(this) {
+            check(callback == null) { "Received an activity start request while another is pending" }
+            callback = onActivityCompleteCallback
+        }
+
+        activityResultLauncher.launch(intent)
+    }
+
+    private fun onActivityComplete() {
+        synchronized(this) {
+            callback?.let { it() }
+            callback = null
+        }
+    }
 }
--- a/android/clientlib-ktx/src/main/java/com/solana/mobilewalletadapter/clientlib/MobileWalletAdapter.kt
+++ b/android/clientlib-ktx/src/main/java/com/solana/mobilewalletadapter/clientlib/MobileWalletAdapter.kt
@@ -14,6 +14,23 @@ import java.util.concurrent.ExecutionException
 import java.util.concurrent.TimeUnit
 import java.util.concurrent.TimeoutException
 
+sealed class TransactionResult<T> {
+    data class Success<T>(
+        val payload: T
+    ): TransactionResult<T>()
+
+    class Failure<T>(
+        val message: String,
+        val e: Exception
+    ): TransactionResult<T>()
+}
+
+/**
+ * Convenience property to access success payload. Will be null if not successful.
+ */
+val <T> TransactionResult<T>.successPayload: T?
+    get() = (this as? TransactionResult.Success)?.payload
+
 class MobileWalletAdapter(
     private val timeout: Int = Scenario.DEFAULT_CLIENT_TIMEOUT_MS,
     private val ioDispatcher: CoroutineDispatcher = Dispatchers.IO
@@ -21,7 +38,7 @@ class MobileWalletAdapter(
 
     private val adapterOperations = LocalAdapterOperations(ioDispatcher)
 
-    suspend fun <T> transact(sender: ActivityResultSender, block: suspend AdapterOperations.() -> T): T = coroutineScope {
+    suspend fun <T> transact(sender: ActivityResultSender, block: suspend AdapterOperations.() -> T): TransactionResult<T> = coroutineScope {
         return@coroutineScope try {
             val scenario = LocalAssociationScenario(timeout)
             val details = scenario.associationDetails()
@@ -36,75 +53,63 @@ class MobileWalletAdapter(
                 }
             }
 
-            withContext(Dispatchers.IO) {
-                val result = try {
+            withContext(ioDispatcher) {
+                try {
                     @Suppress("BlockingMethodInNonBlockingContext")
                     val client = scenario.start().get(ASSOCIATION_TIMEOUT_MS, TimeUnit.MILLISECONDS)
 
                     adapterOperations.client = client
-                    block(adapterOperations)
+                    val result = block(adapterOperations)
+
+                    TransactionResult.Success(result)
                 } catch (e: InterruptedException) {
-                    Log.w(TAG, "Interrupted while waiting for local association to be ready")
-                    throw e
+                    TransactionResult.Failure("Interrupted while waiting for local association to be ready", e)
                 } catch (e: TimeoutException) {
-                    Log.e(TAG, "Timed out waiting for local association to be ready")
-                    throw e
+                    TransactionResult.Failure("Timed out waiting for local association to be ready", e)
                 } catch (e: ExecutionException) {
-                    Log.e(TAG, "Failed establishing local association with wallet", e.cause)
-                    throw e
+                    TransactionResult.Failure("Failed establishing local association with wallet", e)
                 } catch (e: CancellationException) {
-                    Log.e(TAG, "Local association was cancelled before connected", e)
-                    throw e
+                    TransactionResult.Failure("Local association was cancelled before connected", e)
                 } finally {
                     @Suppress("BlockingMethodInNonBlockingContext")
                     scenario.close().get(ASSOCIATION_TIMEOUT_MS, TimeUnit.MILLISECONDS)
                 }
-
-                result
             }
         } catch (e: ExecutionException) {
             when (val cause = e.cause) {
                 is IOException -> {
-                    Log.e(TAG, "IO error while sending operation", cause)
-                    throw e
+                    return@coroutineScope TransactionResult.Failure("IO error while sending operation", cause)
                 }
                 is TimeoutException -> {
-                    Log.e(TAG, "Timed out while waiting for result", cause)
-                    throw e
+                    return@coroutineScope TransactionResult.Failure("Timed out while waiting for result", cause)
                 }
                 is MobileWalletAdapterClient.InvalidPayloadsException -> {
-                    Log.e(TAG, "Transaction payloads invalid", cause)
-                    throw e
+                    return@coroutineScope TransactionResult.Failure("Transaction payloads invalid", cause)
                 }
                 is MobileWalletAdapterClient.NotSubmittedException -> {
-                    Log.e(TAG, "Not all transactions were submitted", cause)
-                    throw e
+                    return@coroutineScope TransactionResult.Failure("Not all transactions were submitted", cause)
                 }
                 is JsonRpc20Client.JsonRpc20RemoteException -> {
-                    when (cause.code) {
-                        ProtocolContract.ERROR_AUTHORIZATION_FAILED -> Log.e(TAG, "Auth token invalid", cause)
-                        ProtocolContract.ERROR_NOT_SIGNED -> Log.e(TAG, "User did not authorize signing", cause)
-                        ProtocolContract.ERROR_TOO_MANY_PAYLOADS -> Log.e(TAG, "Too many payloads to sign", cause)
-                        else -> Log.e(TAG, "Remote exception", cause)
+                    val msg = when (cause.code) {
+                        ProtocolContract.ERROR_AUTHORIZATION_FAILED -> "Auth token invalid"
+                        ProtocolContract.ERROR_NOT_SIGNED -> "User did not authorize signing"
+                        ProtocolContract.ERROR_TOO_MANY_PAYLOADS -> "Too many payloads to sign"
+                        else -> "Remote exception"
                     }
-                    throw e
+                    return@coroutineScope TransactionResult.Failure(msg, cause)
                 }
                 is MobileWalletAdapterClient.InsecureWalletEndpointUriException -> {
-                    Log.e(TAG, "Authorization result contained a non-HTTPS wallet base URI", cause)
-                    throw e
+                    return@coroutineScope TransactionResult.Failure("Authorization result contained a non-HTTPS wallet base URI", cause)
                 }
                 is JsonRpc20Client.JsonRpc20Exception -> {
-                    Log.e(TAG, "JSON-RPC client exception", cause)
-                    throw e
+                    return@coroutineScope TransactionResult.Failure("JSON-RPC client exception", cause)
                 }
-                else -> throw e
+                else -> return@coroutineScope TransactionResult.Failure("Execution exception", e)
             }
         } catch (e: CancellationException) {
-            Log.e(TAG, "Request was cancelled", e)
-            throw e
+            return@coroutineScope TransactionResult.Failure("Request was cancelled", e)
         } catch (e: InterruptedException) {
-            Log.e(TAG, "Request was interrupted", e)
-            throw e
+            return@coroutineScope TransactionResult.Failure("Request was interrupted", e)
         }
     }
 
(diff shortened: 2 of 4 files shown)


android/clientlib-ktx/src/main/java/com/solana/mobilewalletadapter/clientlib/ActivityResultSender.kt
```diff
@@ -1,7 +1,35 @@
...
 
 import android.content.Intent
+import androidx.activity.ComponentActivity
+import androidx.activity.result.ActivityResultLauncher
+import androidx.activity.result.contract.ActivityResultContracts
+import androidx.annotation.GuardedBy
 
-interface ActivityResultSender {
-    fun launch(intent: Intent)
+class ActivityResultSender(
```
creativedrewy, line 9: Consumers will no longer need to implement this as an interface, but instantiate it, providing their app's component activity.

android/clientlib-ktx/src/main/java/com/solana/mobilewalletadapter/clientlib/MobileWalletAdapter.kt
```diff
@@ -7,96 +7,109 @@ import com.solana.mobilewalletadapter.clientlib.scenario.LocalAssociationIntentC
...
 import java.util.concurrent.ExecutionException
 import java.util.concurrent.TimeUnit
 import java.util.concurrent.TimeoutException
 
+sealed class TransactionResult<T> {
+    data class Success<T>(
+        val payload: T
+    ): TransactionResult<T>()
+
+    class Failure<T>(
```
creativedrewy, line 22: Again, we could extend this sealed class type to have specific failure types that most consumers will be concerned with.

examples/example-clientlib-ktx-app/app/src/main/java/com/solanamobile/ktxclientsample/viewmodel/SampleViewModel.kt
```diff
@@ -98,37 +99,39 @@ class SampleViewModel @Inject constructor(
                 }
             }
 
-            persistanceUseCase.persistConnection(
-                currentConn.publicKey,
-                currentConn.accountLabel,
-                currentConn.authToken
-            )
+            result.successPayload?.let { currentConn ->
```
creativedrewy, line 102: This convenience property allows consumers to access the payload without having to create an if/then or when block handling all types of the sealed class.

android/clientlib-ktx/src/main/java/com/solana/mobilewalletadapter/clientlib/ActivityResultSender.kt
Funkatronics: I am not against this change necessarily, but this will break current implementations right? would there be any value to leaving the interface, and providing the implantation here? That would likely be more confusing in the long run but worth discussing?

android/clientlib-ktx/src/main/java/com/solana/mobilewalletadapter/clientlib/MobileWalletAdapter.kt
```diff
@@ -7,96 +7,109 @@ import com.solana.mobilewalletadapter.clientlib.scenario.LocalAssociationIntentC
...
+            val details = scenario.associationDetails()
 
-                val intent = LocalAssociationIntentCreator.createAssociationIntent(details.uriPrefix, details.port, details.session)
-                sender.launch(intent)
+            val intent = LocalAssociationIntentCreator.createAssociationIntent(details.uriPrefix, details.port, details.session)
+            sender.startActivityForResult(intent) {
+                launch {
+                    delay(5000L)
+
+                    Log.v("Andrew", "Cancelling")
```
Funkatronics, line 51: this delay and log... is this test code that should be removed?

android/clientlib-ktx/src/main/java/com/solana/mobilewalletadapter/clientlib/ActivityResultSender.kt
creativedrewy: It only changes the implementation _slightly_ for consumers. They used to have to implement the interface and pass it along. This would have been done in their Activity. Now they just instantiate and pass along. 

We don't have any consumers other than ourselves right now, so I'd rather make the breaking change now than later.

examples/example-clientlib-ktx-app/app/src/main/java/com/solanamobile/ktxclientsample/activity/MainActivity.kt
```diff
@@ -14,23 +13,21 @@ import com.solanamobile.ktxclientsample.ui.theme.KtxClientSampleTheme
 import dagger.hilt.android.AndroidEntryPoint
 
 @AndroidEntryPoint
-class MainActivity : ComponentActivity(), ActivityResultSender {
+class MainActivity : ComponentActivity() {
     override fun onCreate(savedInstanceState: Bundle?) {
         super.onCreate(savedInstanceState)
 
+        val sender = ActivityResultSender(this)
```
Funkatronics, line 20: i like this, easier for the consumer

android/clientlib-ktx/src/main/java/com/solana/mobilewalletadapter/clientlib/MobileWalletAdapter.kt
creativedrewy: OOPS I didn't push

examples/example-clientlib-ktx-app/app/src/main/java/com/solanamobile/ktxclientsample/activity/MainActivity.kt
creativedrewy: Yeah this is the _only_ change required for consumers as of this PR.

android/clientlib-ktx/src/main/java/com/solana/mobilewalletadapter/clientlib/ActivityResultSender.kt
Funkatronics: True true, I dont think its a bad breaking change to make at all. Tho I wonder if we should provide some way for the consumer to provide their own implementation. Do you think this ActivityResultSender implementation will cover all users needs? Maybe it will, but if we think it may need to be tweaked or extended for certain use cases then worth making it an interface with a provided impl IMO.

android/clientlib-ktx/src/main/java/com/solana/mobilewalletadapter/clientlib/MobileWalletAdapter.kt
```diff
@@ -1,102 +1,112 @@
...
+        return@coroutineScope try {
+            val scenario = LocalAssociationScenario(timeout)
+            val details = scenario.associationDetails()
 
-                val intent = LocalAssociationIntentCreator.createAssociationIntent(details.uriPrefix, details.port, details.session)
-                sender.launch(intent)
+            val intent = LocalAssociationIntentCreator.createAssociationIntent(details.uriPrefix, details.port, details.session)
+            sender.startActivityForResult(intent) {
+                launch {
+                    delay(5000L)
```
Funkatronics, line 48: whats the delay for? and should it be configurable?

android/clientlib-ktx/src/main/java/com/solana/mobilewalletadapter/clientlib/ActivityResultSender.kt
creativedrewy: That's a very good question, and probably something we should talk with @sdlaver about.

examples/example-clientlib-ktx-app/app/src/main/java/com/solanamobile/ktxclientsample/activity/MainActivity.kt
Funkatronics: I just want to make sure wee aren't backing into a corner with this. Just want to get steven's perspective so we don't overlook some consumer needs.

android/clientlib-ktx/src/main/java/com/solana/mobilewalletadapter/clientlib/MobileWalletAdapter.kt
creativedrewy: This is ported over from Fakedapp. It's unclear if we want this configurable.

Funkatronics: Okay poked around a bit and I am uncertain if it should be configurable as well haha. We should at least move it to the companion object below like ASSOCIATION_TIMEOUT_MS

android/clientlib-ktx/src/main/java/com/solana/mobilewalletadapter/clientlib/ActivityResultSender.kt
```diff
@@ -1,7 +1,35 @@
...
+            onActivityComplete()
+        }
+
+    fun startActivityForResult(intent: Intent, onActivityCompleteCallback: () -> Unit) {
+        synchronized(this) {
+            check(callback == null) { "Received an activity start request while another is pending" }
+            callback = onActivityCompleteCallback
+        }
+
+        activityResultLauncher.launch(intent)
```
ankur2136, line 26: This needs to be wrapped around with activity handler first to verify if the wallet intent could be handled. 

```
        val packageManager = this@MainActivity.packageManager
        if (intent.resolveActivity(packageManager) != null) {
            startForResult.launch(intent)
        } else {
            lifecycleScope.launch(Dispatchers.Main) {
                Toast.makeText(
                    this@MainActivity,
                    "You need to install a Solana Wallet first (Solflare or Phantom)",
                    Toast.LENGTH_LONG
                ).show()
            }
        }
```

If not then we need to return some sort of object as return for failure (maybe a boolean would suffice)

android/clientlib-ktx/src/main/java/com/solana/mobilewalletadapter/clientlib/MobileWalletAdapter.kt
```diff
@@ -1,102 +1,112 @@
...
-                val details = scenario.associationDetails()
+    suspend fun <T> transact(sender: ActivityResultSender, block: suspend AdapterOperations.() -> T): TransactionResult<T> = coroutineScope {
+        return@coroutineScope try {
+            val scenario = LocalAssociationScenario(timeout)
+            val details = scenario.associationDetails()
 
-                val intent = LocalAssociationIntentCreator.createAssociationIntent(details.uriPrefix, details.port, details.session)
-                sender.launch(intent)
+            val intent = LocalAssociationIntentCreator.createAssociationIntent(details.uriPrefix, details.port, details.session)
+            sender.startActivityForResult(intent) {
```
ankur2136, line 46: Then we need to handle the response for `startActivityForResult` method and return early if there are no activities to handle the intent.

android/clientlib-ktx/src/main/java/com/solana/mobilewalletadapter/clientlib/ActivityResultSender.kt
creativedrewy: At the library level we shouldn't dictate method for notifying the user. BUT - I do think there is a good opportunity either in this library or a companion lib to provide a nice default UI for checking for wallet and presenting some nice default UI. There are two aspects to it:

- This class doesn't crash when there is no destination; returns some kind of status to consumer.
- We create a second, higher level library that has the aforementioned "nice" UI

android/clientlib-ktx/src/main/java/com/solana/mobilewalletadapter/clientlib/MobileWalletAdapter.kt
creativedrewy: Given that this is an instance of the ActivityResultSender library, I think we handle it at that point not here.

android/clientlib-ktx/src/main/java/com/solana/mobilewalletadapter/clientlib/ActivityResultSender.kt
creativedrewy: In general though, this is a good catch; I think this would crash as is

android/clientlib-ktx/src/main/java/com/solana/mobilewalletadapter/clientlib/MobileWalletAdapter.kt
creativedrewy: Delete my previous comment lol. You're 100% correct; this is the place to return early.

android/clientlib-ktx/src/main/java/com/solana/mobilewalletadapter/clientlib/ActivityResultSender.kt
creativedrewy: Yes, it does. Better handling incoming.

creativedrewy:
@ankur2136 @Funkatronics I added library support for the case where no compatible wallet is found. It's a special case error object. Other libraries can perhaps use this to provide convenient default functionality.

ankur2136:
Looks much better.

android/clientlib-ktx/src/main/java/com/solana/mobilewalletadapter/clientlib/ActivityResultSender.kt
sdlaver: We just started publishing to Maven, and we haven't actively promoted this library yet, so there'll never be a better time to make a breaking change.

ankur2136 (approved):

android/clientlib-ktx/src/main/java/com/solana/mobilewalletadapter/clientlib/ActivityResultSender.kt
ankur2136, line 9: Can we also make this final? I don't see a reason for consumers to extend/inherit this

creativedrewy: Very good point

android/clientlib-ktx/src/main/java/com/solana/mobilewalletadapter/clientlib/MobileWalletAdapter.kt
ankur2136: Remove Log.v

creativedrewy: Err, it should be

android/clientlib-ktx/src/main/java/com/solana/mobilewalletadapter/clientlib/ActivityResultSender.kt
creativedrewy: It is, make sure you're looking at latest code @ankur2136

android/clientlib-ktx/src/main/java/com/solana/mobilewalletadapter/clientlib/MobileWalletAdapter.kt
creativedrewy: Just checked and it is; plz make sure you're looking at latest code @ankur2136

android/clientlib-ktx/src/main/java/com/solana/mobilewalletadapter/clientlib/ActivityResultSender.kt
```diff
@@ -1,7 +1,35 @@
...
 import android.content.Intent
+import androidx.activity.ComponentActivity
+import androidx.activity.result.ActivityResultLauncher
+import androidx.activity.result.contract.ActivityResultContracts
+import androidx.annotation.GuardedBy
 
-interface ActivityResultSender {
-    fun launch(intent: Intent)
+class ActivityResultSender(
+    rootActivity: ComponentActivity
```
ankur2136, line 10: this should be renamed to `activity: Activity`

No need to specifically extend `ComponentActivity` for launching the intent.

creativedrewy: Kewl, love it

creativedrewy: Actually, 1ComponentActivity` is required for `registerForActivityResult()` and such

android/clientlib-ktx/src/main/java/com/solana/mobilewalletadapter/clientlib/ActivityResultSender.kt
creativedrewy: I'm trying to make the class final, and Android studio is saying it's redundant

Funkatronics: kotlin classes are final by default

Funkatronics (approved):

ankur2136 (approved):

PR #363, opened 2023-01-23
Wallet Lib Low Power Mode Warning

Funkatronics:
## Summary
- adding a mechanism to alert wallets when the device is in low power mode and incoming MWA connections may be suppressed

## Work Completed
- added new value to MobileWalletAdapterConfig for no connection warning timeout value
- added new LocalWebSocketServerScenario callback that is called if the device is in low power mode, and a connection is not made within the timeout 

### `fakewallet`
- added necessary bits to demonstrate the low power warning and recommended UX (its a crappy alert dialog currently, could be improved on the design) 

### Notes
We have decided to NOT handle the following case in this PR. This case has additional implications that need to be considered and will likely use a different solution in tandem with the logic added in this PR. 

> Sometimes, in low power mode, the connection initially works but subsequent operations on the socket fail (presumably the client side websocket is not immediately paused and remains open for a moment while the wallet app opens, allowing the initial connection to go through and breaking the connection shortly after). In this case, fakewallet will open the authorization page, but any operations (authorize, decline, etc) on this page will hang. 
> 
> 

## Testing
- added 2 LocalWebSocketServerScenario unit tests
- added 6 new `fakewallet` instrumented tests to verify the low power mode behavior (ie shown if low power, not shown otherwise)


android/fakewallet/src/androidTest/res/values/constants.xml
```diff
@@ -0,0 +1,4 @@
+<?xml version="1.0" encoding="utf-8"?>
```
Funkatronics, line 1: this file can be removed, it has no effect (was trying to override the timeout value during testing for faster tests_

Funkatronics:
I am updating this branch with the integration test work so that the tests added (fake wallet tests) here can be run on the CI action

android/walletlib/src/main/java/com/solana/mobilewalletadapter/walletlib/protocol/MobileWalletAdapterConfig.java
```diff
@@ -25,10 +28,12 @@ public class MobileWalletAdapterConfig {
     public final Object[] supportedTransactionVersions;
 
     public MobileWalletAdapterConfig(boolean supportsSignAndSendTransactions,
+                                     @IntRange(from = 0) long noConnectionWarningTimeoutMillis,
```
sdlaver, line 31: Can we append this, rather than add it to the middle? Existing consumers of `walletlib` will see a compilation error, and it'll be easier to interpret when appended. Otherwise, valid values of  `maxTransactionsPerSigningRequest` would now be used for `noConnectionWarningTimeoutMillis`, moving the error to a later parameter that will be harder to interpret.

android/walletlib/src/test/java/android/util/Log.java
```diff
@@ -0,0 +1,29 @@
+package android.util;
+
+/**
+ * Mocks Android's Log class (android.util.Log) so that we can test code that uses the Log
+ * methods without needed to manually mock or inject a logging abstraction for every test.
+ *
+ * This mock is very basic, it does the minimum amount of work to provide a functional test env.
+ */
+public class Log {
```
sdlaver, line 9: What about using Robolectric instead of mocking this class?

android/walletlib/src/test/java/android/os/Looper.java
```diff
@@ -0,0 +1,17 @@
+package android.os;
+
+/**
+ * Mocks Android's Looper class (android.os.Looper) so that we can test code that
+ * uses the Looper abstraction methods without needed to manually mock or inject a Looper
+ * abstraction for every test.
+ *
+ * This mock is very basic, it does the minimum amount of work to provide a functional test env.
+ */
+public class Looper {
```
sdlaver, line 10: What about using Robolectric instead of mocking this class?

android/walletlib/src/main/java/com/solana/mobilewalletadapter/walletlib/scenario/LocalWebSocketServerScenario.java
```diff
@@ -21,15 +23,30 @@ public class LocalWebSocketServerScenario extends Scenario {
...
                                         @NonNull AuthIssuerConfig authIssuerConfig,
-                                        @NonNull Callbacks callbacks,
+                                        @NonNull Scenario.Callbacks callbacks,
                                         @NonNull byte[] associationPublicKey,
                                         @WebSocketsTransportContract.LocalPortRange int port) {
+        this(context, mobileWalletAdapterConfig, authIssuerConfig, callbacks,
+                associationPublicKey, port, new DevicePowerConfigProvider(context));
+    }
+
+    public LocalWebSocketServerScenario(@NonNull Context context,
```
sdlaver, line 38: Similar to comment on `PowerConfigProvider`, I'd make this constructor package-scoped, to limit it's availability to unit tests defined in the same package.

android/walletlib/src/main/java/com/solana/mobilewalletadapter/walletlib/provider/DevicePowerConfigProvider.java
```diff
@@ -0,0 +1,25 @@
+package com.solana.mobilewalletadapter.walletlib.provider;
+
+import android.content.Context;
+import android.os.PowerManager;
+
+import androidx.annotation.Nullable;
+
+public class DevicePowerConfigProvider implements PowerConfigProvider {
```
sdlaver, line 8: Since this is a default implementation, I'd move it into the same package which consumes it, and make it a package-visibility class

android/walletlib/src/main/java/com/solana/mobilewalletadapter/walletlib/scenario/LocalWebSocketServerScenario.java
```diff
@@ -66,9 +87,19 @@ public void onStarted() {
         public void onFatalError() {
             mIoHandler.post(mCallbacks::onScenarioError);
         }
+
+        @Override
+        public void onNoConnectionTimeoutReached() {
+            if (mCallbacks instanceof Callbacks)
```
sdlaver, line 93: NIT: can we always include the braces, even if not syntactically strictly necessary? Even if we avoid a bug once due to this, it's worth the extra characters.

android/walletlib/src/main/java/com/solana/mobilewalletadapter/walletlib/scenario/LocalWebSocketServerScenario.java
```diff
@@ -66,9 +87,19 @@ public void onStarted() {
...
+                ((Callbacks) mCallbacks).onLowPowerAndNoConnectionTimeoutReached();
+        }
     };
 
     private enum State {
         NOT_STARTED, RUNNING, CLOSED
     }
+
+    public interface Callbacks extends Scenario.Callbacks {
+        void onLowPowerAndNoConnectionTimeoutReached();
```
sdlaver, line 103: NIT: the `TimeoutReached` part of this is an implementation detail. Personally, I'd name this `onLowPowerNoConnection`, and leave how the "no connection" state was determined as an implementation detail.

android/fakewallet/src/main/res/values/strings.xml
```diff
@@ -39,4 +39,14 @@
     <string name="str_verification_succeeded">Verification successful</string>
     <string name="str_verification_failed">Verification failed</string>
     <string name="str_verification_not_verifiable">Client not verifiable</string>
+
+    <!-- No connection due to low power warning strings-->
+    <string name="low_power_mode_warning_title">Low Power Mode</string>
+    <string name="str_low_power_mode_warning_dsc">
+        Your device is currently in power saving mode. Incoming wallet connections from your
+        web browser will not work properly when the device battery save mode is active.
```
sdlaver, line 47: Not just from web browser - could be from any source.

android/fakewallet/src/main/res/values/constants.xml
```diff
@@ -0,0 +1,4 @@
+<?xml version="1.0" encoding="utf-8"?>
+<resources>
+    <integer name="no_connection_timeout_millis">3000</integer>
```
sdlaver, line 3: Whether or not integer constants should be defined in XML resources is an open question (and I'm happy to revisit), but IMO we should do it for `fakewallet` as a whole. Right now, there is already a companion object in `MobileWalletAdapterViewModel` with constants defined; let's keep it all centralized there.

android/walletlib/src/main/java/com/solana/mobilewalletadapter/walletlib/protocol/MobileWalletAdapterConfig.java
```diff
@@ -13,6 +13,9 @@ public class MobileWalletAdapterConfig {
 
     public final boolean supportsSignAndSendTransactions;
 
+    @IntRange(from = 0)
+    public final long noConnectionWarningTimeoutMillis;
```
sdlaver, line 17: NIT: elsewhere in `walletlib` and `clientlib` we use the suffix `Ms`, rather than `Millis`

android/fakewallet/src/main/java/com/solana/mobilewalletadapter/fakewallet/MobileWalletAdapterActivity.kt
```diff
@@ -31,6 +32,23 @@ class MobileWalletAdapterActivity : AppCompatActivity() {
...
+                viewModel.lowPower.collect {
+                    // should use dialog fragment, etc. but this is a quick demo
+                    AlertDialog.Builder(this@MobileWalletAdapterActivity)
+                        .setTitle(R.string.low_power_mode_warning_title)
+                        .setMessage(R.string.str_low_power_mode_warning_dsc)
+                        .setPositiveButton(R.string.label_return_to_dapp) { _, _ ->
+                            Log.w(TAG, "Connection failed due to device low power mode, returning to dapp.")
+                            finish()
+                        }
+                        .setNeutralButton(R.string.label_close_dialog) { d, _ -> d.dismiss() }
```
sdlaver, line 46: Do we really want a neutral button here? Real wallets can be more complex if they desire, but I'd propose we only implement the straightforward case for `fakewallet`, which in this instance is the positive button you've defined above.

android/fakewallet/src/main/java/com/solana/mobilewalletadapter/fakewallet/MobileWalletAdapterActivity.kt
```diff
@@ -31,6 +32,23 @@ class MobileWalletAdapterActivity : AppCompatActivity() {
...
         }
 
+        lifecycleScope.launch {
+            repeatOnLifecycle(Lifecycle.State.CREATED) {
+                viewModel.lowPower.collect {
+                    // should use dialog fragment, etc. but this is a quick demo
+                    AlertDialog.Builder(this@MobileWalletAdapterActivity)
+                        .setTitle(R.string.low_power_mode_warning_title)
+                        .setMessage(R.string.str_low_power_mode_warning_dsc)
+                        .setPositiveButton(R.string.label_return_to_dapp) { _, _ ->
```
sdlaver, line 42: MAYBE: use the standard [android.R.string.ok](https://developer.android.com/reference/android/R.string#ok) here

android/fakewallet/src/main/java/com/solana/mobilewalletadapter/fakewallet/MobileWalletAdapterViewModel.kt
```diff
@@ -32,6 +28,9 @@ class MobileWalletAdapterViewModel(application: Application) : AndroidViewModel(
     val mobileWalletAdapterServiceEvents =
         _mobileWalletAdapterServiceEvents.asSharedFlow() // expose as event stream, rather than a stateful object
 
+    private val _lowPower = MutableSharedFlow<Boolean>(1)
```
sdlaver, line 31: Rather than create a new flow for this, could you create another subclass of `MobileWalletAdapterServiceRequest`, representing the low-power/no-connection event?

android/fakewallet/src/androidTest/java/com/solana/mobilewalletadapter/fakewallet/MainActivityTest.kt
```diff
@@ -0,0 +1,258 @@
...
+import org.junit.Before
+import org.junit.Rule
+import org.junit.Test
+import org.junit.runner.RunWith
+import java.util.concurrent.CountDownLatch
+import java.util.concurrent.TimeUnit
+
+@RunWith(AndroidJUnit4::class)
+@LargeTest
+class MainActivityTest {
```
sdlaver, line 36: Should this class have a `@Before` method which resets `TestScopeLowPowerMode` to a known state before each test is run? That way, only tests that need it in the non-default state need to explicitly configure it. Otherwise, the behavior of this test may vary depending on the order of test execution.

android/walletlib/src/main/java/com/solana/mobilewalletadapter/walletlib/provider/PowerConfigProvider.java
```diff
@@ -0,0 +1,5 @@
+package com.solana.mobilewalletadapter.walletlib.provider;
+
+public interface PowerConfigProvider {
```
sdlaver, line 3: Proposal: move this class to `com.solana.mobilewalletadapter.walletlib.scenario`, and make it package-scoped. That way, we limit its use to only test code.

sdlaver (changes requested):
Proposal: what if we were to extract an interface from `Scenario`, and define a new base class, `LocalScenarioBase`. All scenarios which are local to the device would extend from there. Then, we'd move the timeout logic out of the `WebSocketServer` class and into `LocalScenarioBase`. This would ensure that, no matter what transport a scenario makes use of, we have common local timeout logic.

Funkatronics:
@sdlaver I am still seeing the issue where the connection initially does work (we receive a valid auth request) but the web socket connection quietly closes. I am not seeing a way on our end to detect this case explicitly.  I think this state could occur for any request, as the low power state could change at any time. I propose that we add the connection timeout any time we try to complete a request on low power.

I have added this change in my latest commit [26c3261](https://github.com/solana-mobile/mobile-wallet-adapter/pull/363/commits/26c32615f8de67ae29d937f46a43e8f8bf4e49eb). This implementation is handling this case properly for authorization request. If you agree that this is needed I can add some additional automated tests for all of the request types.

Funkatronics:
these darn tests! I don't see why the changes in this PR would make the fakewallet tests flake out on CI - they are passing when I run them locally. 

We should probably add screenshot functionality to the test solution so we can more easily debug these failures, but I don't want to do that in this PR.

android/walletlib/src/main/java/com/solana/mobilewalletadapter/walletlib/scenario/LocalScenario.java
```diff
@@ -0,0 +1,403 @@
...
+            mNoConnectionTimeoutHandler = mTimeoutExecutorService.schedule(() -> {
+                Log.i(TAG, "No connection timeout reached");
+                if (mCallbacks instanceof Callbacks) {
+                    ((Callbacks) mCallbacks).onLowPowerAndNoConnection();
+                }
+            }, noConnectionTimeout, TimeUnit.MILLISECONDS);
+    }
+
+    private void stopNoConnectionTimer() {
+        if (mNoConnectionTimeoutHandler != null) mNoConnectionTimeoutHandler.cancel(true);
```
sdlaver, line 139: Should we also null out the `mNoConnectionTimeoutHandler` here, so it can be garbage collected?

android/walletlib/src/main/java/com/solana/mobilewalletadapter/walletlib/scenario/LocalScenario.java
```diff
@@ -0,0 +1,403 @@
...
+                        final String authToken = mAuthRepository.toAuthToken(authRecord);
+                        request.complete(new MobileWalletAdapterServer.AuthorizationResult(
+                                authToken, authorize.publicKey, authorize.accountLabel,
+                                authorize.walletUriBase));
+                    } else {
+                        request.completeExceptionally(new MobileWalletAdapterServer.RequestDeclinedException(
+                                "authorize request declined"));
+                    }
+
+                    startNoConnectionTimer();
```
sdlaver, line 220: Curious: what necessitates starting the no connection timer here (and elsewhere below, in response to `MobileWalletAdapterServer.MethodHandlers` callbacks)?

sdlaver (changes requested):

android/fakewallet/src/main/java/com/solana/mobilewalletadapter/fakewallet/MobileWalletAdapterViewModel.kt
```diff
@@ -478,11 +475,18 @@ class MobileWalletAdapterViewModel(application: Application) : AndroidViewModel(
             Log.d(TAG, "'${event.identityName}' deauthorized")
             event.complete()
         }
+
+        override fun onLowPowerAndNoConnection() {
+            viewModelScope.launch {
```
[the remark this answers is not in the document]
Funkatronics: ooo good call

android/walletlib/src/main/java/com/solana/mobilewalletadapter/walletlib/scenario/DevicePowerConfigProvider.java
```diff
@@ -0,0 +1,25 @@
+package com.solana.mobilewalletadapter.walletlib.scenario;
+
+import android.content.Context;
+import android.os.PowerManager;
+
+import androidx.annotation.Nullable;
+
+class DevicePowerConfigProvider implements PowerConfigProvider {
```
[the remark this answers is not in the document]
Funkatronics: ah darnit, forgot that here. Wonder if I can automate that...

android/walletlib/src/main/java/com/solana/mobilewalletadapter/walletlib/scenario/LocalScenario.java
```diff
@@ -0,0 +1,403 @@
...
+import java.util.concurrent.ScheduledExecutorService;
+import java.util.concurrent.ScheduledFuture;
+import java.util.concurrent.TimeUnit;
+import java.util.concurrent.atomic.AtomicInteger;
+
+public abstract class LocalScenario implements Scenario {
+    private static final String TAG = LocalScenario.class.getSimpleName();
+
+    @NonNull
+    public byte[] associationPublicKey;
```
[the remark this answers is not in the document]
Funkatronics: Nope, just a mistake. at one point this got moved into the interface and I didnt revert it back correctly

android/walletlib/src/main/java/com/solana/mobilewalletadapter/walletlib/scenario/LocalScenario.java
```diff
@@ -0,0 +1,403 @@
...
+
+    @Override
+    public void start() {
+        startNoConnectionTimer();
+    }
+
+    @Override
+    public abstract void close();
+
+    public Long getNoConnectionTimeout() {
```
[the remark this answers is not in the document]
Funkatronics: nope good catch!

android/walletlib/src/main/java/com/solana/mobilewalletadapter/walletlib/scenario/LocalScenario.java
Funkatronics: did you see [this comment](https://github.com/solana-mobile/mobile-wallet-adapter/pull/363#issuecomment-1430769129)?

I can reproduce this case where the initial connection succeeds and we get a valid request (even tho we are on low power mode) but subsequent operations (like, user confirming authorization) will hang. 

I have logged the web socket server to see if we receive _anything_ after the initial request is received that would tell us that the connection is closed/paused and I am not seeing anything.

Funkatronics: I can make a video that shows this if youd like, but steps to repo are the same:
- have device in low power
- open dapp and initiate a request 
- if you:
  - **do** see the request land, try to complete it (tap authorize), the request completion will hang
  - **do not** see the connection request, you should see the low power warning. try these steps again to see the other case

On saga I am seeing the first case more often than the second so I think its important to fix.

android/walletlib/src/main/java/com/solana/mobilewalletadapter/walletlib/scenario/LocalScenario.java
```diff
@@ -0,0 +1,403 @@
...
+    public MessageReceiver createMessageReceiver() {
+        return new MobileWalletAdapterSession(
+                this,
+                new MobileWalletAdapterServer(mMobileWalletAdapterConfig, mIoLooper, mMethodHandlers),
+                mSessionStateCallbacks);
+    }
+
+    @Override
+    public void start() {
+        startNoConnectionTimer();
```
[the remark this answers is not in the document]
Funkatronics: My thinking was that the executor service used for the timeout scheduling will manage its own thread context and trigger the callback from that context, but I am overlooking to potential for concurrent modification of these objects. 

I think the onLowPowerAndNoConnection callback should be posted into the io looper too so its not triggered from a non IO thread.

android/walletlib/src/main/java/com/solana/mobilewalletadapter/walletlib/scenario/LocalScenario.java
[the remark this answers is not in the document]
sdlaver: Super minor, but no need to use a boxed type `Long` here; primitive `long` will do. I suspect that this would be optimized away at build time, but 🤷 .

android/walletlib/src/main/java/com/solana/mobilewalletadapter/walletlib/scenario/LocalScenario.java
```diff
@@ -0,0 +1,406 @@
...
+
+    @Nullable
+    private ScheduledFuture<?> mNoConnectionTimeoutHandler;
+    private final ScheduledExecutorService mTimeoutExecutorService =
+            Executors.newSingleThreadScheduledExecutor();
+
+    protected LocalScenario(@NonNull Context context,
+                            @NonNull MobileWalletAdapterConfig mobileWalletAdapterConfig,
+                            @NonNull AuthIssuerConfig authIssuerConfig,
+                            @NonNull Scenario.Callbacks callbacks,
```
sdlaver, line 69: WDYT about `LocalAssociationUri.createScenario` taking a `LocalScenario.Callbacks` instead of `Scenario.Callbacks`? That would allow this to be specifically typed, and we wouldn't need the type check on line 132.

android/walletlib/src/main/java/com/solana/mobilewalletadapter/walletlib/scenario/LocalScenario.java
sdlaver: What I'm worried about is that, once the session is open, there is no requirement on how long between actions the dApp is allowed to take. Let me think about this a little...

android/walletlib/src/main/java/com/solana/mobilewalletadapter/walletlib/scenario/LocalScenario.java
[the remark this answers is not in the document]
Funkatronics: dang I actually forgot about that distinction in Java. Kotlin made my java skills rusty lol

android/walletlib/src/main/java/com/solana/mobilewalletadapter/walletlib/scenario/LocalScenario.java
Funkatronics: I tried that actually but because this overrides the `createScenario` method from `AssociationUri` I cant put a subtype of `Scenerio.Callbacks` there.

We could add an overload for `createScenario` to `AssociationUri` that takes `LocalScenario.callbacks` but I think that's not the best approach. Do you know of another way to resolve this?

android/walletlib/src/main/java/com/solana/mobilewalletadapter/walletlib/scenario/LocalWebSocketServerScenario.java
```diff
@@ -13,23 +13,38 @@
...
-public class LocalWebSocketServerScenario extends Scenario {
+public class LocalWebSocketServerScenario extends LocalScenario {
     @WebSocketsTransportContract.LocalPortRange
     public final int port;
 
     @NonNull
     private final LocalWebSocketServer mWebSocketServer;
     private State mState = State.NOT_STARTED;
 
+    private PowerConfigProvider mPowerManager;
```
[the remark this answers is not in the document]
Funkatronics: ah good catch, left over from old impl.

sdlaver (approved):

PR #388, opened 2023-03-07
Make clientlib-ktx and fakedapp wait for the completion of the previous request

### LINKED ISSUE #94: If you trigger an association activity while the last one is still concluding, it will silently fail

### LINKED ISSUE #367: [clientlib-ktx] Race condition when making subsequent launch calls from MWA
In clientlib-ktx, if the `transact` method errors in any way, the activity call may not resolve correctly. Calling into MWA again after that rightfully causes a crash. Clientlib-ktx should handle this gracefully.

sdlaver:
Fixes #94
Fixes #367

### DIFF
--- a/android/clientlib-ktx/src/main/java/com/solana/mobilewalletadapter/clientlib/ActivityResultSender.kt
+++ b/android/clientlib-ktx/src/main/java/com/solana/mobilewalletadapter/clientlib/ActivityResultSender.kt
@@ -5,12 +5,12 @@ import android.content.Intent
 import androidx.activity.ComponentActivity
 import androidx.activity.result.ActivityResultLauncher
 import androidx.activity.result.contract.ActivityResultContracts
-import androidx.annotation.GuardedBy
+import androidx.lifecycle.whenResumed
+import kotlinx.coroutines.*
 
 class ActivityResultSender(
-    rootActivity: ComponentActivity
+    private val rootActivity: ComponentActivity
 ) {
-    @GuardedBy("this")
     private var callback: (() -> Unit)? = null
 
     private val activityResultLauncher: ActivityResultLauncher<Intent> =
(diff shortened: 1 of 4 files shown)


ankur2136:
Tested with store code and works perfectly.

ankur2136 (approved):

Funkatronics:
lgtm, love the use of lifecycle events

Funkatronics (approved):

PR #407, opened 2023-03-09
FakeDApp offchain message signing should use non-random messages

sdlaver:
This improves both repeatability of tests, as well as comprehension of signing in wallets


sdlaver:
@nickcruz I'd propose this change instead of #233.

PR #408, opened 2023-03-09
Create a demo bundled MWA session that performs three actions

sdlaver:
Actions:
1. Authorize
2. Sign a message indicating ownership of account
3. Delay for 1.5s (simulating presenting the signed message to a server, and receiving some server-generated transaction object)
4. Sign a transaction and send it to the network

This allows verifying that wallets (including FakeWallet) handle this sequence of operations correctly.

### DIFF
--- a/android/fakedapp/src/main/res/values/strings.xml
+++ b/android/fakedapp/src/main/res/values/strings.xml
@@ -20,7 +20,8 @@
     <string name="label_sign_txn_x1">Sign txn x1</string>
     <string name="label_sign_txn_x3">x3</string>
     <string name="label_sign_txn_x20">x20</string>
-    <string name="label_authorize_sign_txn">Combined authorize and sign txn x1</string>
+    <string name="label_authorize_sign_txn">Authorize + Txn</string>
+    <string name="label_authorize_sign_msg_txn">Authorize + Msg + Send Txn</string>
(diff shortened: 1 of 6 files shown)


android/fakedapp/src/main/res/values/strings.xml
```diff
@@ -20,7 +20,8 @@
     <string name="label_sign_txn_x1">Sign txn x1</string>
     <string name="label_sign_txn_x3">x3</string>
     <string name="label_sign_txn_x20">x20</string>
-    <string name="label_authorize_sign_txn">Combined authorize and sign txn x1</string>
+    <string name="label_authorize_sign_txn">Authorize + Txn</string>
```
Funkatronics, line 23: NIT: should this be "Authorize + Sign Txn" or is that too big for the available space?

Funkatronics:
should we add an automated test for this case with fakewallet?

android/fakedapp/src/main/res/values/strings.xml
sdlaver: Yeah, it was running right up against the available space. That screen is getting pretty crowded!

sdlaver:
> should we add an automated test for this case with fakewallet?

Added

Funkatronics (approved):

PR #412, opened 2023-03-12
Update spec for reauthorize

dimosmera:
Removes identity as a possible parameter for reauthorize

Tested with the latest ^1.0.0 version

### DIFF
--- a/spec/spec.md
+++ b/spec/spec.md
@@ -430,18 +430,12 @@ reauthorize
 
 ```
 {
-    “identity”: {
-        “uri”: “<dapp_uri>”,
-        “icon”: “<dapp_icon_relative_path>”,
-        “name”: “<dapp_name>”,
-    },
     “auth_token”: “<auth_token>”,
 }
 ```
 
 where:
 
-- `identity`: as defined for [`authorize`](#authorize)
 - `auth_token`: an opaque string previously returned by a call to [`authorize`](#authorize), [`reauthorize`](#reauthorize), or [`clone_authorization`](#clone_authorization)
 
 ###### Result


steveluscher:
Hmm, were we just taking about this, @sdlaver? Is this actually supposed to be part of the spec, but we’re not enforcing it in the JS reference implementation?

sdlaver:
> Hmm, were we just taking about this, @sdlaver? Is this actually supposed to be part of the spec, but we’re not enforcing it in the JS reference implementation?

I think it's a good idea to merge `authorize` and `reauthorize` into a single call. It originally served a distinct purpose, but as the spec developed that purpose became narrower and narrower.

That said, just dropping the identity from `reauthorize` like this is not a good idea. Wallets should be given as much access as possible to dApp identity info, and if a dApp fails to present identical information on a reauth, it should probably be rejected out of hand.

@dimosmera would you mind sharing the motivation behind this PR?

steveluscher:
Yup. 

https://github.com/solana-mobile/mobile-wallet-adapter/blob/20f9a571bb55656fae79ce344c36ca1311c0c651/js/packages/mobile-wallet-adapter-protocol/src/types.ts#L81

ankur2136:
> Yup.
> 
> https://github.com/solana-mobile/mobile-wallet-adapter/blob/20f9a571bb55656fae79ce344c36ca1311c0c651/js/packages/mobile-wallet-adapter-protocol/src/types.ts#L81

java/kotlin lib uses all params

https://github.com/solana-mobile/mobile-wallet-adapter/blob/main/android/clientlib/src/main/java/com/solana/mobilewalletadapter/clientlib/protocol/MobileWalletAdapterClient.java#L262-L287

steveluscher:
Hey, you don’t have to rub it in :)

dimosmera:
I see. Thank you both. Indeed that was my motivation but I was only looking at the JavaScript implementation

PR #416, opened 2023-03-14
Add `identity` to the method signature of `reauthorize`

steveluscher:
## Summary

This part of the JS implementation was not in compliance with the spec. This probably explains bad behavior some have been noting with certain wallets. If a wallet is particularly strict about receiving an `identity` through `reauthorize`, we have _not_ been sending it. This would cause `ERROR_AUTHORIZATION_FAILED` or `-32602` in all cases.

## Change

In this PR, we update the type to include the `identity` param. This is a breaking change for clients using `mobile-wallet-adapter-protocol` or `mobile-wallet-adapter-protocol-web3js` that aren't already supplying it.

## Test plan

Tested sending multiple transactions with the web and React Native sample apps.

### DIFF
--- a/js/packages/mobile-wallet-adapter-protocol/src/types.ts
+++ b/js/packages/mobile-wallet-adapter-protocol/src/types.ts
@@ -78,7 +78,7 @@ export interface GetCapabilitiesAPI {
     >;
 }
 export interface ReauthorizeAPI {
-    reauthorize(params: { auth_token: AuthToken }): Promise<AuthorizationResult>;
+    reauthorize(params: { auth_token: AuthToken; identity: AppIdentity }): Promise<AuthorizationResult>;
 }
 export interface SignMessagesAPI {
     signMessages(params: {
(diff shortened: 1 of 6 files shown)


Funkatronics (commented):
Code looks good, but want to double check with others on the impact of this change before we merge

js/packages/mobile-wallet-adapter-protocol/src/types.ts
```diff
@@ -78,7 +78,7 @@ export interface GetCapabilitiesAPI {
     >;
 }
 export interface ReauthorizeAPI {
-    reauthorize(params: { auth_token: AuthToken }): Promise<AuthorizationResult>;
+    reauthorize(params: { auth_token: AuthToken; identity: AppIdentity }): Promise<AuthorizationResult>;
```
Funkatronics, line 81: can we make `identity` optional here?

Funkatronics (changes requested):

js/packages/mobile-wallet-adapter-protocol/src/types.ts
steveluscher: We chatted about this. I'm opposed (when dApps don't identify themselves, it leads to a bad UX), and in any case the `identity` _object_ is not optional, so to make it optional here would require a change to the spec.

PR #434, opened 2023-04-20
Add docsite references

Michaelsulistio:
https://docs.solanamobile.com is live so I am updating the Github documentation to point to resources that can be found on the doc site. This is part of the process of rolling out the doc site, and will slowly start to designate the doc site as the single source of truth.

In this PR:
- Add references to docsite in the main `README.md`
- Add a reference to our doc site in `integration_guide.md`
- Add a reference to our doc site in `@solana-mobile/mobile-wallet-adapter-protocol` `README.md`
- Add a sample app `README.md` that outlines sample apps and refers to docsite

Let me know if there's other places you can think of to refer to the doc site!

### DIFF
--- a/README.md
+++ b/README.md
@@ -7,9 +7,19 @@
 
 _Part of the [Solana Mobile Stack](https://github.com/solana-mobile/solana-mobile-stack-sdk)_
 
-See our [documentation website](https://docs.solanamobile.com/)
-
 Join us on [Discord](https://discord.gg/solanamobile)
+
+## See our [documentation website](https://docs.solanamobile.com/) for:
+- [General Development Setup](https://docs.solanamobile.com/getting-started/development_setup)
+- React Native
+    - [Quickstart Setup](https://docs.solanamobile.com/react-native/quickstart)
+    - [dApp Integration Guide](https://docs.solanamobile.com/react-native/mwa_integration_rn)
+    - [Hello World Tutorial](https://docs.solanamobile.com/getting-started/hello_world_tutorial)
+- Android
+    - [Quickstart Setup](https://docs.solanamobile.com/android-native/quickstart)
+    - [dApp Integration Guide](https://docs.solanamobile.com/android-native/mwa_integration)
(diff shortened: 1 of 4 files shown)


README.md
```diff
@@ -9,6 +9,18 @@ _Part of the [Solana Mobile Stack](https://github.com/solana-mobile/solana-mobil
...
+## See our [documentation website](https://docs.solanamobile.com/) for:
+- [General Development Setup](https://docs.solanamobile.com/getting-started/development_setup)
+- React Native
+    - [Quickstart Setup](https://docs.solanamobile.com/react-native/quickstart)
+    - [dApp Integration Guide](https://docs.solanamobile.com/react-native/mwa_integration_rn)
+    - [Hello World Tutorial](https://docs.solanamobile.com/getting-started/hello_world_tutorial)
+- Android
+    - [Quickstart Setup](https://docs.solanamobile.com/android-native/quickstart)
+    - [dApp Integration Guide](https://docs.solanamobile.com/android-native/mwa_integration)
+- [Sample App Reference](https://docs.solanamobile.com/sample-apps/sample_app_overview)
```
creativedrewy, line 21: nit: do we want to be consistent on calling them "dApp" vs "app"? I personally don't care actually, but maybe others have strong opinions

creativedrewy (approved):

README.md
Michaelsulistio: that's a good point and we should be standardizing the language across all our docs. I can update to `dApp` in this PR and make docsite changes too.

On a similar note, I was discussing with @Funkatronics was the naming between `dApp` vs `DApp` and we landed on `dApp`.

Michaelsulistio: Deciding to leave these as `dApp` instead of `app`. I think the term `dApp` is useful in making a distinction between dApps and Wallet Apps. This is a distinction that's made in the [MWA protocol spec](https://solana-mobile.github.io/mobile-wallet-adapter/spec/spec.html#dapp-identity-verification) and its useful.

PR #443, opened 2023-04-28
[WIP] Add react native walletlib package

Michaelsulistio:


examples/example-react-native-wallet/screens/MobileWalletAdapterEntrypointBottomSheet.tsx
```diff
@@ -0,0 +1,109 @@
...
+    case MobileWalletAdapterServiceEventType.SignTransactions:
+    case MobileWalletAdapterServiceEventType.SignMessages:
+      return <SignPayloadsScreen request={request as SignPayloadsRequest} />;
+    case MobileWalletAdapterServiceEventType.AuthorizeDapp:
+      return <AuthenticationScreen request={request as AuthorizeDappRequest} />;
+    default:
+      return <ActivityIndicator size="large" />;
+  }
+}
+
```
Michaelsulistio, line 41: This is the MWA Entrypoint component. It is what will be rendered when the user intent is accepted by the wallet app.

examples/example-react-native-wallet/screens/MobileWalletAdapterEntrypointBottomSheet.tsx
```diff
@@ -0,0 +1,109 @@
...
+  SignAndSendTransactionsRequest,
+} from '@solana-mobile/mobile-wallet-adapter-walletlib';
+
+import AuthenticationScreen from '../bottomsheets/AuthenticationScreen';
+import LoadingScreen from '../bottomsheets/LoadingScreen';
+import SignAndSendTransactionsScreen from '../bottomsheets/SignAndSendTransactionsScreen';
+import SignPayloadsScreen from '../bottomsheets/SignPayloadsScreen';
+import WalletProvider from '../components/WalletProvider';
+
+function getRequestScreenComponent(
```
Michaelsulistio, line 22: Each request is handled with the appropriate Bottomsheet Screen UI

js/packages/mobile-wallet-adapter-walletlib/android/src/main/java/com/solanamobile/mobilewalletadapterwalletlib/reactnative/MobileWalletAdapterBottomSheetActivity.kt
```diff
@@ -0,0 +1,55 @@
...
+
+import android.os.Bundle
+import android.view.Gravity
+import android.view.WindowManager
+
+import com.facebook.react.ReactActivity
+import com.facebook.react.ReactActivityDelegate
+import com.facebook.react.ReactRootView
+
+class MobileWalletAdapterBottomSheetActivity : ReactActivity() {
```
Michaelsulistio, line 11: This is the custom Bottom Sheet activity that we've included in the library and it's styled to support floating window

js/packages/mobile-wallet-adapter-walletlib/android/src/main/java/com/solanamobile/mobilewalletadapterwalletlib/reactnative/MobileWalletAdapterBottomSheetActivity.kt
```diff
@@ -0,0 +1,55 @@
...
+
+    /**
+     * Returns the name of the main component registered from JavaScript. This is used to schedule
+     * rendering of the component.
+     */
+    override protected  fun getMainComponentName() = "MobileWalletAdapterEntrypoint"
```
Michaelsulistio, line 17: This corresponds to the component that we register in the RN app's `index.js`

js/packages/mobile-wallet-adapter-walletlib/android/src/main/java/com/solanamobile/mobilewalletadapterwalletlib/reactnative/SolanaMobileWalletAdapterWalletLibModule.kt
```diff
@@ -0,0 +1,386 @@
...
+        }
+    }
+
+    private fun sendEvent(reactContext: ReactContext, eventName: String, params: WritableMap? = null) {
+        reactContext
+            .getJSModule(DeviceEventManagerModule.RCTDeviceEventEmitter::class.java)
+            .emit(eventName, params)
+    }
+
+    private inner class MobileWalletAdapterScenarioCallbacks : LocalScenario.Callbacks {
```
Michaelsulistio, line 325: In these scenario callbacks, we receive requests from the DApp and simply send them over to javascript.

js/packages/mobile-wallet-adapter-walletlib/android/src/main/java/com/solanamobile/mobilewalletadapterwalletlib/reactnative/SolanaMobileWalletAdapterWalletLibModule.kt
```diff
@@ -0,0 +1,386 @@
...
+        sealed class SignPayloads(override val request: SignPayloadsRequest) : MobileWalletAdapterRemoteRequest(request)
+        data class SignTransactions(override val request: SignTransactionsRequest) : SignPayloads(request)
+        data class SignMessages(override val request: SignMessagesRequest) : SignPayloads(request)
+        data class SignAndSendTransactions(
+            override val request: SignAndSendTransactionsRequest,
+            val endpointUri: Uri,
+        ) : MobileWalletAdapterRemoteRequest(request)
+    }
+
+    private var request: MobileWalletAdapterServiceRequest? = null
```
Michaelsulistio, line 38: setting the request field will trigger sending the request to react

js/packages/mobile-wallet-adapter-walletlib/android/src/main/java/com/solanamobile/mobilewalletadapterwalletlib/reactnative/SolanaMobileWalletAdapterWalletLibModule.kt
```diff
@@ -0,0 +1,386 @@
...
+            return
+        } else if (associationUri !is LocalAssociationUri) {
+            Log.e(TAG, "Current implementation of fakewallet does not support remote clients")
+            return
+        }
+
+        // created a scenario, told it to start (kicks off some threads in the background)
+        // we've kept a reference to it in the global state of this module (scenario)
+        // this won't be garbage collected and will just run, sit & wait for an incoming connection
+        scenario = associationUri.createScenario(
```
Funkatronics, line 90: new scenario gets created each time. An edge case would be if react native caller tries to create a new scenario when one already exists. Should we return an error or do anything different in this case? Maybe we should rename the method to make it more clear that it will always overriding any previous scenario

js/packages/mobile-wallet-adapter-walletlib/android/src/main/java/com/solanamobile/mobilewalletadapterwalletlib/reactnative/SolanaMobileWalletAdapterWalletLibModule.kt
```diff
@@ -0,0 +1,386 @@
...
+    // Converts a react "ReadableArray" into a Kotlin ByteArray.
+    // Expects ReadableArray to be an Array of ints, where each int represents a byte.
+    private fun convertFromReactByteArray(reactByteArray: ReadableArray): ByteArray {
+        return ByteArray(reactByteArray.size()) { index ->
+            reactByteArray.getInt(index).toByte()
+        }
+    }
+
+    @ReactMethod
+    fun log(message: String) {
```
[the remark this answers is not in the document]
Michaelsulistio: No, definitely need to clean this up. There's a bunch of other small things I'll need to clean up too.

Michaelsulistio:
@steveluscher Requesting feedback for the topics below:




- The structure/schema of the requests in [requests.ts](https://github.com/solana-mobile/mobile-wallet-adapter/pull/443/files/e6957bfd2b4918b6494dc7ccd93f871159a48fa9#diff-151e73f46fabc2326a45f10f4b10dcb4425b390039c355b87237f25ce837ba0b)
- useMobileWalletAdapterRequest in [index.ts](https://github.com/solana-mobile/mobile-wallet-adapter/pull/443/files/e6957bfd2b4918b6494dc7ccd93f871159a48fa9#diff-ba09f0133f3896a2d716ec4df785c1020a2d7f30b919ed9599425891e244fa2f). This is the approach I ended up choosing (a stream of events -> convert to request classes). The users would subscribe to MWA requests by calling this custom hook.
- Any improvements in the [bottom sheet entrypoint modal component](https://github.com/solana-mobile/mobile-wallet-adapter/pull/443/files/e6957bfd2b4918b6494dc7ccd93f871159a48fa9#diff-4483ac955cd071d0e130fa23ba4312acb44cd18d5bf4da142c06b7e2283b22cf)
- Any improvements in the bottomsheet request screen components (ie [AuthorizationScreen](https://github.com/solana-mobile/mobile-wallet-adapter/pull/443/files/e6957bfd2b4918b6494dc7ccd93f871159a48fa9#diff-ec1002b150590d9f4c5d2e6e6a11bcc800a73b5e3b0bfd2ae672a21bb038e9f9))

Michaelsulistio:
@Funkatronics Had a cases I wanted to discuss, that might affect the JS side of things.

1. Distinguishing MWA Remote Requests from MWA Events
 We currently have a singular stream of MWA requests coming from `useMobileWalletAdapterRequest`. These requests will be `AuthorizeDappRequest` or `SignMessagesRequest`, etc. These are typed as `MobileWalletAdapterServiceRemoteRequest`.

We also have another set of "events" (not requests) that might need to be communicated to JS. These are the Scenario lifecycle events (`onScenarioReady`, `onScenarioServingClients`, `onScenarioServingComplete`, etc). Should we create a separate JS type/stream for these types of events? e.g. `useMobileWalletAdapterServiceEvents`. This way, it won't "override" the current `MobileWalletAdapterServiceRemoteRequest`. (ie. We're on the Auth Screen, then a `onLowPowerAndNoConnection` request comes in and causes our UI to change).

js/packages/mobile-wallet-adapter-walletlib/android/src/main/java/com/solanamobile/mobilewalletadapterwalletlib/reactnative/SolanaMobileWalletAdapterWalletLibModule.kt
Michaelsulistio: So, this method is never actually called by the JS dev right now (at least we don't explicitly expose it). JS side solely interacts with this under-the-hood when they call `useMobileWalletAdapterRequest`.

Should we allow the JS side to explicitly call this?

js/packages/mobile-wallet-adapter-walletlib/android/src/main/java/com/solanamobile/mobilewalletadapterwalletlib/reactnative/SolanaMobileWalletAdapterWalletLibModule.kt
```diff
@@ -0,0 +1,386 @@
...
+            AuthIssuerConfig(walletName),
+            MobileWalletAdapterScenarioCallbacks()
+        ).also { it.start() }
+
+        Log.d(TAG, "scenario created: $walletName")
+    }
+
+
+    @ReactMethod
+    fun closeScenario() {
```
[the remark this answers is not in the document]
Michaelsulistio: yup this should be removed.

js/packages/mobile-wallet-adapter-walletlib/android/src/main/java/com/solanamobile/mobilewalletadapterwalletlib/reactnative/SolanaMobileWalletAdapterWalletLibModule.kt
```diff
@@ -0,0 +1,386 @@
...
+    }
+
+    @ReactMethod
+    fun completeWithAuthorize(publicKey: ReadableArray, accountLabel: String?, walletUriBase: String?, authorizationScope: ReadableArray?) {
+        Log.d(TAG, "completeWithAuthorize: authorized public key = $publicKey")
+        (request as? MobileWalletAdapterServiceRequest.AuthorizeDapp)?.request?.let { authRequest ->
+            authRequest.completeWithAuthorize(
+                convertFromReactByteArray(publicKey),
+                accountLabel,
+                null, // walletUriBase,
```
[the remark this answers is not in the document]
Michaelsulistio: yup. I had some trouble understanding what this Auth Scope byte array should look like and how to generate it.

steveluscher (dismissed):
This is such exciting work. You're making something genuinely new that's going to be explosively useful!

I was in the middle of making a bunch of targeted feedback, but I think I have to zoom _way_ out and start a conversation about the architecture.

**Here are the things that concern me the most**:

 * **We rely on the React render cycle for event handling** \
    Your app handles changes in the session event under processing by watching for state changes in effect hooks. This is a sure fire way to both delay processing, and miss events. When you have an event in hand, just respond to it directly; don't send it on a round trip through `setState` and `useEffect`.
* **There's no correlation between an MWA request and the response** \
    This code presumes that one event will be received and that, when the wallet calls `completeWith*()`, the request that spawned that event will be waiting for the response. We can't guarantee this at all.
* **We're relying on React to listen for URL intents** \
    I'm pretty sure that the Android side should listen for the session initialization event, not the JS side.

**Here are the architectural solutions I see to these problems**. Caveat: I haven't thought about them nearly as long as you.

1. The native module, and the native module alone, should listen for session initialization intents (the `solana-wallet:` URL)
2. The JS side receive nothing but an event emitter, and should do no more than listen for events:
    * `onSessionStart`
    * `onSessionEstablished`
    * `onAuthorize`
    * `onReauthorize`
    * `onSignMessages`
    * `onSignTransactions`
    * `onSignAndSendTransactions`
    * `onSessionComplete`
    *  `onSessionError`
    * `onSessionTerminated`
    * `onSessionLowPowerNoConnection`
3. Each event _must_ carry a request that has two correlation ids; one for the session and one for the event itself:
    ```ts
    enum MWARequestType {
      SignMessagesRequest,
      SignTransactionsRequest,
      /* ... */
    }
    interface IMWARequest {
      __type: MWARequestType;
      requestId: string;
      sessionId: string;
    };
    type SignMessagesRequest = Readonly<{
      __type: MWARequestType.SignMessagesRequest,
      payloads: Uint8Array[],
      /* ... */
    }> & IMWARequest;
    type SignTransactionsRequest = Readonly<{
      __type: MWARequestType.SignTransactionsRequest,
      cluster: string,
      payloads: Uint8Array[],
      /* ... */
    }> & IMWARequest
    type MWARequest = (SignMessagesRequest | SignTransactionsRequest)
    ```
4. Responding to the request is a matter of passing in the original event and the typed response. The native side correlates the response to the original session/request using the correlation ids. If the session has gone away, no problem. ([Typescript playground](https://www.typescriptlang.org/play?#code/CYUwxgNghgTiAEkoGdnwMoHtoDsoFlMAjASwhAHUoJyAXAQWCgAdaQYqaRaAZEo+AG8AUPHjJaUWiTDxMOAEohk2AG4gAFCHU5aALiHjlyEvICSwAxJgkcAcwA08OAEcArstoWrtG-fgAvk5wyMzyyCAGbjgA1jiYAO44AJQGqpgkwMIBwsIgOG4AtvD4FPRK7p4AKgCezAgiYugkdjj4xlB2yhUeEg6iGC04VTBQOMhQYNLhPZ79YgD0AFTwAHTr8EsL2cK2bDAAZpMIZqXlIJUSQgMA+je0dZElZbMStfUA3AOuvV6W4r5bHYvmIIqhTDhvAC-MDsl8HvVBq12qhOt0Lr94ABeeBKKDAeQQGoAHka8DuCKeZ1etHeIFWzWRHS6yBp83gzCgNQgmHxyAMAFU9gAOegwUY1ADaAF12cs1hsttkAHzwABk8FOLwxnnhjyRw1G40m03GNOxuJA+MJJLJFMeBmpOrejwZQxGYwmUwhrOdtHZkDcEnYPhh7M53N5wH58CFulF4q5MrlK3Wq022wCqo1WvOl1owkpzzzmJxGkZbWZ6Pz8AAPgaPcbvTM-clckWKyiJiylKFwggcXiCTgiaTxEMQMAu2iY3HaAmJTLAsq9YiK42vabfX3xgPLdaR7bx61JxuTT7BSKxYvpcvVwgnTuIhbO1Xt2Fd3WG0bNz7ex+Ii+YQDmiZscGcZQ1E0H5PAMV9UR7P1gmUACnng7t0SfEBUngdJMi+ECcDAiCVAgdQNBgiQ4PdH9zxbfNkKw6jWjPMD337HC8OAAjQNNEioIov1HW1BiSNQqJYniJJOIyYBrlBBISFoMAAAt4EE-NVntepknksREBQB8RN+Ok3SZBDq1+PQBn0+AFgWeAAHlWAhagiQ5OAAFpmBgTAwGMeBaBUhAQlQpwEhUmQ1JINApPgHAZBAdzKWAVYbP0rBcAIYgyEoNzuEYFh9k4Og+CIVZ5F7ATKP9MT+3gFADWnRCsOSEFbPgIg4CgGJ2rEMBDOLGlTPXWi2JpayOrshznNNNyag8kBvN8-zUEC4K6t3cLItU+AYvixJ4sS5LHlS9KmmwMZstIcgSoKphWHYO7eH4CrFEgsjoKQzbn0a0bPTos0UI4vr9K6q1eoGHIcmEEIBI0AoaAatBojiRJwL+oZmsszxkicQRjxwSdsZjJcAja6anIAaVhj7yIRtwkca1GpIxtB-qbLcaTxwwTBPYBWK3AwyYp+zqdyOHPo0BmmZRyT0eR78AfG1t8cJ4m32F29yY+SmAFEFAURyFAMQoYsKKRVMnH6QGEIA))
    ```ts
    SolanaMobileWalletAdapterWalletLib.resolve(request, {signedPayloads: /* ... */});
    ```
5. All your React code does is to listen for events.
    ```ts
    const [session, setSession] = useState();
    const handleMWAEvent = useCallback((request: MWARequest) => {
      switch(request.__type) {
        case MWARequestType.SessionStart:
          if (session && session.id !== request.sessionId) {
            // Oh no, we already have a session going on that's not this one!
          }
          setSession({id: request.sessionId, state: 'starting'});
          break;
        case MWARequestType.SignMessages:
          // Queue up the sign messages UI
          break;
        /* ... */
      }
    }, []);
    useMobileWalletAdapterProtocol(handleMWAEvent);
    ```

I'll stop here for thoughts. I know that was a lot! I essentially think that the way to go here is to create a regular RPC-style interface between the client (the JS) and the server (the Android native side) and to establish correlation ids between requests & responses, a type contract that governs what those should be, and only really two methods to talk over the bridge (subscribe/publish).

js/packages/mobile-wallet-adapter-walletlib/src/index.ts
```diff
@@ -0,0 +1,202 @@
...
+        handleServiceRequest(typedRequest);
+    });
+
+    // Listens for session events
+    eventEmitter.addListener(MOBILE_WALLET_ADAPTER_SESSION_EVENT_BRIDGE_NAME, (newEvent) => {
+        const typedSessionEvent: MobileWalletAdapterSessionEvent | null = bridgeEventToMWASessionEvent(newEvent);
+        handleSessionEvent(typedSessionEvent);
+    });
+    return eventEmitter;
+}
```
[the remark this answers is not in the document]
Michaelsulistio: Ah gotcha. I think what happened here was I left this `export` by mistake. Definitely should be private,

js/packages/mobile-wallet-adapter-walletlib/src/index.ts
```diff
@@ -0,0 +1,202 @@
...
+    maxMessagesPerSigningRequest: number;
+    supportedTransactionVersions: Array<TransactionVersion>;
+    noConnectionWarningTimeoutMs: number;
+}
+
+export function subscribeToNativeEvents(
+    handleServiceRequest: (event: MobileWalletAdapterServiceRequest | null) => void,
+    handleSessionEvent: (event: MobileWalletAdapterSessionEvent | null) => void,
+) {
+    const eventEmitter = new NativeEventEmitter();
```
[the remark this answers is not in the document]
Michaelsulistio: Actually, I had this initially but I kept getting some console warning about missing `removeAllListeners` function.

steveluscher:
Another thought. Something else you could do would be to vend an `AsyncIterator` when a session starts. That iterator would vend requests until the session closes, demand that they be responded to in series, and would throw if it dies.

```ts
useMobileWalletAdapterSession(async (sessionId: string, mwaRequests: AsyncIterator<MWARequest>) => {
  setSession({id: sessionId, state: 'starting'});
  try {
    await for (const mwaRequest in mwaRequests) {
      if (session.state === 'starting') {
        setState({id: sessionId, state: 'started'});
      }
      switch(mwaRequest.__type) {
        case 'SignTransactions':
          // Sign 'em.
          mwaRequest.respond(signedTransactions);
      }
    }
    setSession({id: sessionId, state: 'complete'});
  } catch (e) {
    setSession({id: sessionId, state: 'failed'});
  }
});

```

Michaelsulistio:
@steveluscher Thanks for the feedback. I'll address the architectural problems you mentioned:

**We rely on the React render cycle for event handling**
Gotcha. What I'm understanding here, is that my API right now is one level of abstraction higher than needed. We need to remove one layer of `setState/useEffect` and let the user respond directly to the event without it going through a hook. 

I'm having trouble understanding how the UI would update and how users can take an action to `respond` to the request.

In my head the flow would be:
- There would still need to some state held in the component to know which UI to show like:
```
const {request, setRequest} = setState(null)
// or
const {screen, setScreen} = setState(null)
```
- The user then calls the state setter this in the callback handler which triggers UI update
- Now, on the `SignTransaction` screen, we pass the request as a prop and can call `request.resolve()` or `resolve(request)`.

**There's no correlation between an MWA request and the response**
This was something @Funkatronics and I discussed and we're planning on solving incrementally. We had some early ideas of tagging each request with some `uid` like your requestId idea. Definitely seems important to add in the initial release of this library.

I didn't consider also having a sessionId as I was under the assumption that we're working with a single session throughout the duration of the bottomsheet activity.

Could multi-session support  be added as a follow up? And we just warn users that starting multiple sessions will break behavior? When would a user need multi-session support?

This code presumes that one event will be received and that, when the wallet calls completeWith*(), the request that spawned that event will be waiting for the response. We can't guarantee this at all.

**We're relying on React to listen for URL intents**
Unless I'm misunderstanding here, we're actually relying on the Android native code to listen for intents. In my walletlib package I have a custom activity `MobileWalletAdapterBottomSheetActivity` that has intent filters for `solana-wallet`. Its also styled to be compatible for bottom sheet rendering over another app.

Funkatronics:
Seriously kick ass effort here @Michaelsulistio and thank you @steveluscher for the comprehensive review. 

We are going to try to break out much of the feedback here into separate tickets/issues, identify which issues can be backlogged and which need to go in for a "v0.1" release, and start hammering away. The goal for v0.1 is to provide something that is functional, with a hopefully mostly stable API. Further under the hood improvements can be chipped away as needed/as capacity allows.

steveluscher:
**About using the render cycle for event handling**: you never want to handle events by bouncing them off state because:

* Not all events beget rendering. Consider `reauthorize`. That's 100% behind the scenes; no UI involved
* _You do not control the render cycle_. React can preempt it, cancel it, run it twice for fun, suspend it. You _need_ one event to correspond to one synchronous event handler.

Wherever an event should trigger UI changes (eg. a screen transition) then your event handler, as appropriate, can do the normal thing and cause that transition, as appropriate for the app's navigation framework. We should not be opinionated about what that is, nor should we presume that the way a wallet transitions from screen to screen is by reading a state variable in a render method.

**About correlating sessions**: if you use an event emitter pattern, you're going to have a difficult time tracking which session a given event belongs to unless you track session starts and completes/failures _perfectly_. A correlation id for the session makes it so that you don't have to be perfect. Nothing to do with concurrent sessions. If you use something more like the `AsyncIterator` pattern above, then you can probably forego this, since you can guarantee that a session starts at the top of that function, and dies by the end of it; sort of like the `transact()` method in the MWA protocol JS implementation.

**About correlating requests and responses**: there's nothing stopping a client from calling multiple methods on the MWA server without waiting for a response. I can totally do this:
```ts
await transact(wallet => {
  await wallet.authorize(...);
  await Promise.all([
    wallet.signMessage(...);
    wallet.signAndSendTransactions(...);
    wallet.signTransaction(...);
  ]);
});
```

You could imagine a wallet that wants to collect all of those for display in one UI, handle them in parallel, and return them to the _correct_ caller. To support that, you'd have to implement the event emitter architecture and let wallets field all the events as they come. The `AsyncIterator` pattern would force the wallet to process the method calls in some serial order, but would obviate the need for correlation ids. Tradeoffs.

> …we're actually relying on the Android native code to listen for intents

I just saw `Linking.getInitialUrl()` in the code and a call to `createScenario()` from the JS side in `initializeScenario()`. I'm not sure the JS side should kick off the scenario at all. The native side should listen for the intent, create the scenario, and then just fire events across the bridge. The only things you have to watch for are:

* the native side shouldn't do anything unless it's [tracking](https://reactnative.dev/docs/native-modules-android#sending-events-to-javascript) at least one JavaScript listener.
* you have to figure out some way to communicate the app identity (name, icon, url) to the native side; probably through configuration upfront

Funkatronics (approved):
@steveluscher Mike and I have chatted on this, we are going to break the requested changes out into separate issues and tackle them together. I am approving this PR as is with this caveat

PR #478, opened 2023-06-03
Native Module Serialization Refactor

Funkatronics:


Michaelsulistio (approved):
very clean, I like this change.

How does the Serializable pattern handle errors/malformed JSON? Are the errors more helpful?

Michaelsulistio (approved):
very clean, I like this change.

How does the Serializable pattern handle errors/malformed JSON? Are the errors more helpful?

Funkatronics:
> very clean, I like this change.
> 
> How does the Serializable pattern handle errors/malformed JSON? Are the errors more helpful?

@Michaelsulistio yes, it will tell you what part of the json was unexpected, and print the input json. I have not had any trouble debugging serialization errors. And of course we can directly print the json strings, and the json objects so it fairly easy to debug when things go wrong. 

Also, Kotlin serialization is multiplatform. We could use the exact same serialization on the JS side to ensure that everything lines up. We could even provide all of the request/response types in a separate kotlin multiplatform lib so that both android and JS can pull form the exact same types and serialization implementation. I did not go this rout now because JSON.stringify() on the JS side has been working fine. But we have room for further cleanup down the road if things start to get more complex with future iterations to the MWA spec.

PR #481, opened 2023-06-11
Update README.md

jnwng:
also noticed that this link was broken! but fixed the typo anyway

wasn't sure what was a suitable replacement for this link


Funkatronics:
Thanks @jnwng! I will take a look at the link

Funkatronics:
Yeah you are right, we don't currently have another suitable link for that text. I swapped out the link with something that I think is relevant for that section. 

Will let @Michaelsulistio make the final call when he is back this week.

PR #505, opened 2023-07-21
Expose Deauthorize Requests/Event to React Native

Funkatronics:


js/packages/mobile-wallet-adapter-walletlib/android/src/main/java/com/solanamobile/mobilewalletadapterwalletlib/reactnative/JsonSerializationUtils.kt
```diff
@@ -76,9 +77,10 @@ internal object MobileWalletAdapterResponseSerializer : JsonContentPolymorphicSe
     override fun selectDeserializer(element: JsonElement): DeserializationStrategy<out MobileWalletAdapterResponse> =
         if ((element as? JsonObject)?.containsKey("failReason") == true) FailReasonTransformingSerializer
         else if ((element as? JsonObject)?.containsKey("publicKey") == true) AuthorizeDappResponse.serializer()
```
Michaelsulistio, line 79: Ideally, we also include a `__type` in the response object, so instead of checking for unique fields we can just check a more readable `__type` field. Not a big deal/blocker though and probably not worth a whole refactor.

Michaelsulistio (approved):
Looks good! After these changes, there's a couple of TODOs left that we synced about offline.

1. Move `initializeScenario` into `useEffect` to avoid redundant, expensive operations.
2. On JS side, fix the redundant `authorizationScope` field in`AuthorizationRequest`.

js/packages/mobile-wallet-adapter-walletlib/android/src/main/java/com/solanamobile/mobilewalletadapterwalletlib/reactnative/JsonSerializationUtils.kt
Funkatronics: yeah I actually started to add that but ran into some issues on the JS side due to the use of types vs classes. I couldn't find a clean way to enforce the __type field so it was getting omitted if I did not explicitly add it to every response, which seemed like a highly error prone situation if we are relying on those types for deserialization. Obviously there has to be a better way to handle this but decided to leave it out for a separate PR.

PR #507, opened 2023-07-25
[WIP] Upgrade ktx MWA to facilitate auth/reauth automatically

creativedrewy:


Funkatronics:
I like it. `connect` aligns with the walet-standard standard features which is the plan for the next MWA spec version. this should map nicely to that new spec, and provides an upgrade in the meantime.

Funkatronics (approved):

Issue #440, opened 2023-04-26
Align MWA API with Wallet Standard


Funkatronics:

Funkatronics:
completed in #521 #562 and  #563, spec in near final

Issue #438, opened 2023-04-26
Multi Account Support


Funkatronics:

d-reader-josip:
Is there any wallet out there which actually allows users to authorize multiple addresses?

Did I understand this right: it's up to wallets to allow users to select multiple addresses to authorize (include within the auth tokens scope), but it's up to dApps to implement the addressSelector. So far no mobile wallet implemented the addresses-to-authorize selector so there is no point for dApps to implement the addressSelector (since it will always be a single wallet address authorized per auth token).

Somewhat related comment from my previous feature request: https://github.com/solana-mobile/mobile-wallet-adapter/issues/328#issuecomment-1500234382
Steven Luschers answere on SM discord: https://discord.com/channels/988649555283308564/1092846107542950038/1094140237506166885

He also says that multi-account feature is gtg for the most part?
Why did you add this task then: " Implement Multiple Accounts feature in Mobile Wallet Adapter"

Funkatronics:
Hey @d-reader-josip, 

The mobile wallet adapter spec allows wallets to return a list of accounts that have been authorized for the scope ([see here](https://github.com/solana-mobile/mobile-wallet-adapter/issues/44)). But currently this functionality is not fleshed out through the rest of the protocol, and is not used in our example apps (see issue #44 for more background and info about this). So we want to, at the very least, update the rest of the privileged methods in the spec to add an `accounts` field to specify which account(s) the dapp would like the wallet to use for that request. Additionally, if we add some kind of account feature flag or profiling, dapps and wallets could use this multi account api to handle more complex interactions like accounting for multi sig wallets, using burners or payer accounts, issue read only accounts alongside a writable account, etc. This is an evolution of the spec that will allow for more interesting use cases and interactions, while still supporting todays common use case of 1 account per session.  

> Is there any wallet out there which actually allows users to authorize multiple addresses?
> it's up to wallets to allow users to select multiple addresses to authorize (include within the auth tokens scope), but it's up to dApps to implement the addressSelector. 
> there is no point for dApps to implement the addressSelector
How these features are used in practice remains up to the dapps and wallets. Dapps and Wallets can choose how they want to adopt and handle these new features. As long as they are compatible with the underlying spec, the UI is up to them. I can see possible use cases that require some kind of address selector ui in the dapp, and others that allow the dapp to select the correct account based on its features without user input.

Issue #319, opened 2022-10-27
Support CAIP-2-style identifiers in the Mobile Wallet Adapter spec


sdlaver:
This will align MWA with wallet-standard

PR #521, opened 2023-08-02
Mwa 2.0 spec

Funkatronics:
- [x] #437 
- [x] #319 
- [x] #438 
- [x] #439 
- [x] #440

TBDs:

- Encrypt/Decrypt
  - this feature needs more research, and with the addition of the feature description functionality in this spec encryption can be added as an experimental feature shortly after the 2.0 spec is finalized and published 
- Ephemeral Accounts
  - this feature will be added as an optional feature extension along with the release of the 2.0 spec.

### DIFF
--- a/spec/spec.md
+++ b/spec/spec.md
@@ -131,6 +131,7 @@ These features are optional, wallet endpoints may choose to implement these feat
 
 - [`solana:signAndSendTransaction`](#sign_and_send_transactions), an optional Wallet RPC method. 
 - `solana:signInWithSolana`, an optional extension to the [`authorize`](#authorize) method. 
+- [`solana:cloneAuthorization`](#clone_authorization)
 
 ## Transport
 
@@ -538,7 +539,6 @@ get_capabilities
 
 ```
 {
-    "supports_clone_authorization": <supports_clone_authorization>,
     "max_transactions_per_request": <max_transactions_per_request>,
     "max_messages_per_request": <max_messages_per_request>,
     "supported_transaction_versions": [<supported_transaction_versions>, ...]
@@ -548,7 +548,6 @@ get_capabilities
 
 where:
 
-- `supports_clone_authorization`: `true` if the [`clone_authorization`](#clone_authorization) method is supported, otherwise `false`
 - `max_transactions_per_request`: (optional) if present, the max number of transaction payloads which can be signed by a single [`sign_transactions`](#sign_transactions) or [`sign_and_send_transactions`](#sign_and_send_transactions) request. If absent, the implementation doesn't publish a specific limit for this parameter.
 - `max_messages_per_request`: (optional) if present, the max number of transaction payloads which can be signed by a single [`sign_messages`](#sign_messages) request. If absent, the implementation doesn't publish a specific limit for this parameter.
 - `supported_transaction_versions`: the Solana network transaction formats supported by this wallet endpoint. Allowed values are those defined for [`TransactionVersion`](https://solana-labs.github.io/solana-web3.js/modules.html#TransactionVersion) (for e.g., `"legacy"`, `0`, etc).
--- a/spec/spec.md
+++ b/spec/spec.md
@@ -129,8 +129,8 @@ These features are mandatory and must be implemented by wallet endpoints. Dapp e
 
 These features are optional, wallet endpoints may choose to implement these features. Dapp endpoints can check if a wallet supports these features by calling [`get_capabilities`](#get_capabilities).  
 
-- [`solana:signAndSendTransaction`](#sign_and_send_transactions)
-- [`solana:signInWithSolana`](#authorize)
+- [`solana:signAndSendTransaction`](#sign_and_send_transactions), an optional Wallet RPC method. 
+- `solana:signInWithSolana`, an optional extension to the [`authorize`](#authorize) method. 
 
 ## Transport
 
@@ -522,7 +522,7 @@ where:
 - `max_transactions_per_request`: (optional) if present, the max number of transaction payloads which can be signed by a single [`sign_transactions`](#sign_transactions) or [`sign_and_send_transactions`](#sign_and_send_transactions) request. If absent, the implementation doesn't publish a specific limit for this parameter.
 - `max_messages_per_request`: (optional) if present, the max number of transaction payloads which can be signed by a single [`sign_messages`](#sign_messages) request. If absent, the implementation doesn't publish a specific limit for this parameter.
 - `supported_transaction_versions`: the Solana network transaction formats supported by this wallet endpoint. Allowed values are those defined for [`TransactionVersion`](https://solana-labs.github.io/solana-web3.js/modules.html#TransactionVersion) (for e.g., `"legacy"`, `0`, etc).
-- `features`: a list of [feature identifiers](#feature-identifiers) for the optional features supported by this wallet endpoint.
+- `features`: a list of [feature identifiers](#feature-identifiers) for the [optional features](#optional-features) supported by this wallet endpoint. [Mandatory features](#mandatory-features) should not be included in this list because all valid wallet endpoints must implement these features. Dapp endpoints can assume that mandatory features are supported by the wallet. 
 
 ###### Errors
 {: .no_toc }
--- a/spec/spec.md
+++ b/spec/spec.md
@@ -122,6 +122,7 @@ Feature identifiers are used to identify features that are supported by a wallet
 - `solana:signMessage`
 - `solana:signTransaction`
 - `solana:signAndSendTransaction`
+- `solana:signInWithSolana`
 
 ## Transport
 
@@ -354,7 +355,7 @@ where:
   - `icon`: (optional) a relative path (from `uri`) to an image asset file of an icon identifying the dapp endpoint making this authorization request
   - `name`: (optional) the display name for this dapp endpoint
 - `chain`: (optional) if set, the [chain identifier](#chain-identifiers) for the chain with which the dapp endpoint intends to interact; supported values include `solana:mainnet`, `solana:testnet`, `solana:devnet`. If not set, defaults to `solana:mainnet`.
-- `auth_token`: (optional) an opaque string previously returned by a call to [`authorize`](#authorize), or [`clone_authorization`](#clone_authorization). When present, the wallet endpoint should attempt to reauthorize the dapp endpoint sliently without promting the user. 
+- `auth_token`: (optional) an opaque string previously returned by a call to [`authorize`](#authorize), or [`clone_authorization`](#clone_authorization). When present, the wallet endpoint should attempt to reauthorize the dapp endpoint silently without prompting the user. 
 
 ###### Result
 {: .no_toc }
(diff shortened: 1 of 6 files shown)


spec/spec.md
```diff
@@ -748,6 +723,98 @@ where:
 
 The wallet endpoint should present the provided messages for approval. If approved, the wallet endpoint should sign the messages with the private key for the authorized account address, and return the signed messages to the dapp endpoint. The signatures should be appended to the message, in the same order as `addresses`.
 
+#### sign_in_with_solana
```
Funkatronics, line 726: I think this should be moved to a non-privileged method. The idea with SIWS (in wallet-standard) is that it combines connect + sign message and returns the account that was used to sign the payload (along with the signature). MWA differs from web-wallets in that we have a session that authorizes a set of accounts for use within the session. I think Sign in With Solana should effectively be a way to establish an authorized session with a single account.

spec/spec.md
```diff
@@ -102,6 +103,82 @@ _reflector_ - an intermediary which brokers connections between two endpoints, w
...
+### Feature Identifiers
+
+Feature identifiers are used to identify features that are supported by a wallet or account. The currently defined standard features include: 
+
+- `solana:signMessages`
+- `solana:signTransactions`
+- `solana:signAndSendTransaction`
+- `solana:signInWithSolana`
+
+#### Feature Descriptions
```
Funkatronics, line 127: This section is TBD, and where I need the most help/input.

How should we describe feature interfaces in a language agnostic way? Here I have drafted a spec using JSON RPC method descriptions similar to how we are already describing methods in the Wallet RPC Interface section.  Questions:

1. Should features be versioned? this would allow us to increment specific features/methods without updating the entire spec. 
2. How should params, results and errors be described? Should we simply provide the names for these and expect endpoints to consult relevant documentation for the exact types and formats of these fields? Or should feature descriptions include complete documentation of the fields including their types, descriptions, etc? We could use an existing IDL standard for this, such as [Apache Avro](https://avro.apache.org/docs/1.11.1/specification/), or define our own. 

Motivation:
We want wallet endpoints to be able to describe their capabilities in a way that allows folks to extend the spec without requiring an updated version of the core specification, similar to how the wallet-standard allows new and experimental features to be defined and described by the [`Wallet` Interface](https://github.com/wallet-standard/wallet-standard/blob/ce280f05f6eb16d0e77939827e0b7fa294460e1f/packages/core/base/src/wallet.ts#L33)

spec/spec.md
```diff
@@ -360,17 +446,29 @@ where:
...
+This method allows the dapp endpoint to request authorization from the wallet endpoint for access to [privileged methods](#privileged-methods). On success, it returns an `auth_token` providing access to privileged methods, along with addresses and optional labels for all authorized accounts. It may also return a URI suitable for future use as an [endpoint-specific URI](#endpoint-specific-uris). After a successful call to [`authorize`](#authorize)`, the current session will be placed into an authorized state, with privileges associated with the returned `auth_token`. On failure, the current session with be placed into the unauthorized state.
```
Funkatronics, line 453: make sure this covers the case where chain is changed.

spec/spec.md
```diff
@@ -102,6 +103,82 @@ _reflector_ - an intermediary which brokers connections between two endpoints, w
...
+
+Namespaced identifiers are used in the format `${namespace}:${reference}` to refer to objects like blockchains and features in a canonical and human readable form. 
+
+### Chain Identifiers
+
+Namespaced chain identifiers are used to refer to blockchains with which a dapp intends to interact with. This scheme is compatible with [CAIP-2](https://github.com/ChainAgnostic/CAIPs/blob/master/CAIPs/caip-2.md) chain IDs, but are not required to be used.
+
+The currently supported Solana network chains are `solana:mainnet`, `solana:testnet`, and `solana:devnet`. 
+
+Previous versions of this specification used a `cluster` parameter to specify the Solana network cluster with which the dapp endpoint intends to interact. The supported values for the `cluster` parameter were `mainnet-beta`, `testnet`, and `devnet`. These values have been replaced by the above chain references.
```
Funkatronics, line 116: add backwards compatibility for previous clusters, as well as authorize

spec/spec.md
```diff
@@ -23,6 +23,7 @@ This specification uses [semantic versioning](https://en.wikipedia.org/wiki/Soft
 | Version | Description                                                                                                                         |
 | ------- | ----------------------------------------------------------------------------------------------------------------------------------- |
 | 1.0.0   | Initial release version of the Mobile Wallet Adapter specification (identical to pre-release version 0.9.1)                         |
+| 2.0.0   | Initial draft of Mobile Wallet Adaper 2.0                                                                                           |
```
sdlaver, line 26: NIT: we should only formally rev the version on the `main` branch when this is no longer in draft

spec/spec.md
```diff
@@ -102,6 +103,27 @@ _reflector_ - an intermediary which brokers connections between two endpoints, w
...
+
+Namespaced identifiers are used in the format `${namespace}:${reference}` to refer to objects like blockchains and features in a canonical and human readable form. 
+
+### Chain Identifiers
+
+Namespaced chain identifiers are used to refer to blockchains with which a dapp intends to interact with. This scheme is compatible with [CAIP-2](https://github.com/ChainAgnostic/CAIPs/blob/master/CAIPs/caip-2.md) chain IDs, but are not required to be used.
+
+The currently supported Solana network chains are `solana:mainnet`, `solana:testnet`, and `solana:devnet`. 
+
+Previous versions of this specification used a `cluster` parameter to specify the Solana network cluster with which the dapp endpoint intends to interact. The supported values for the `cluster` parameter were `mainnet-beta`, `testnet`, and `devnet`. These values have been replaced by the above chain references, but will continue to be supported in the interest of backwards compatibility with pervious versions of the specification.
```
sdlaver, line 116: NIT: `pervious` -> `previous`

spec/spec.md
sdlaver, line 1: As separately discussed, @Funkatronics to document forward and backward compability considerations with this spec revision

spec/spec.md
```diff
@@ -331,7 +356,10 @@ where:
...
   - `name`: (optional) the display name for this dapp endpoint
-- `cluster`: (optional) if set, the Solana network cluster with which the dapp endpoint intends to interact; supported values include `mainnet-beta`, `testnet`, `devnet`. If not set, defaults to `mainnet-beta`.
+- `chain`: (optional) if set, the [chain identifier](#chain-identifiers) for the chain with which the dapp endpoint intends to interact; supported values include `solana:mainnet`, `solana:testnet`, `solana:devnet`. If not set, defaults to `solana:mainnet`.
+- `auth_token`: (optional) an opaque string previously returned by a call to [`authorize`](#authorize), or [`clone_authorization`](#clone_authorization). When present, the wallet endpoint should attempt to reauthorize the dapp endpoint silently without prompting the user. 
+- `sign_in_payload`: (optional) a value object containing the payload portion of a [Sign In With Solana message](https://siws.web3auth.io/spec). If present, the wallet endpoint should present the SIWS message to the user and return the resulting signature and signed message, as described below.
```
sdlaver, line 361: Will all the `solana:` namespaced features be mandatory? Now that we have feature identifiers, we should document which ones are mandatory, and which are optional. For anything optional, we should call out in the spec the sections that relate to that feature.

spec/spec.md
```diff
@@ -331,7 +356,10 @@ where:
...
+- `chain`: (optional) if set, the [chain identifier](#chain-identifiers) for the chain with which the dapp endpoint intends to interact; supported values include `solana:mainnet`, `solana:testnet`, `solana:devnet`. If not set, defaults to `solana:mainnet`.
+- `auth_token`: (optional) an opaque string previously returned by a call to [`authorize`](#authorize), or [`clone_authorization`](#clone_authorization). When present, the wallet endpoint should attempt to reauthorize the dapp endpoint silently without prompting the user. 
+- `sign_in_payload`: (optional) a value object containing the payload portion of a [Sign In With Solana message](https://siws.web3auth.io/spec). If present, the wallet endpoint should present the SIWS message to the user and return the resulting signature and signed message, as described below.  
+- `cluster`: (deprecated) (optional) if set, the Solana network cluster with which the dapp endpoint intends to interact; supported values include `mainnet-beta`, `testnet`, `devnet`. This parameter is deprecated but maintained for backwards compatibility with previous versions of the spec, and will be ignored if the `chain` parameter is present.
```
sdlaver, line 362: MAYBE: to simplify this, what if we make `cluster` an alias for chain, and if both are present, `cluster` is ignored? They can then take the same values (both new and legacy chain descriptors), which would simplify implementation logic

spec/spec.md
```diff
@@ -340,149 +368,126 @@ where:
 {
     “auth_token”: “<auth_token>”,
     “accounts”: [
-        {“address”: “<address>", “label”: “<label>”},
+        {
+            “address”: “<address>", 
+            “label”: “<label>”, 
+            "chains": ["<chain_id>", ...],
```
sdlaver, line 374: How would a multi-chain account function in the remainder of this spec? The other methods don't include a chain specifier, so there wouldn't be any way to use an alternate chain.

spec/spec.md
sdlaver, line 361: Also, did we align our payload for this with Wallet Standard?

spec/spec.md
```diff
@@ -340,149 +368,126 @@ where:
...
 - `wallet_uri_base`: (optional) if this wallet endpoint has an [endpoint-specific URI](#endpoint-specific-uris) that the dapp endpoint should use for subsequent connections, this member will be included in the result object. The dapp endpoint should use this URI for all subsequent connections where it expects to use this `auth_token`.
+- `sign_in_result`: (optional) if the authorize request included a [Sign In With Solana](https://siws.web3auth.io/spec) sign in payload, the result must be returned here as a value object containing the following:
+  - `address`: the address of the account that was signed in. The address of the account may be different from the provided input address, but must be the address of one of the accounts returned in the `accounts` field. 
+  - `signed_message`: the base64-encoded signed message payload
+  - `signature`: the base64-encoded signature
+  - `signature_type`: (optional) the type of the message signature produced. If not provided, the signature must be `"ed25519"`.
```
sdlaver, line 402: > If not provided
Provided in the request, or in the response?

spec/spec.md
```diff
@@ -340,149 +368,126 @@ where:
...
+- `ERROR_CHAIN_NOT_SUPPORTED` if the wallet endpoint does not support the requested chain
 
-The returned `auth_token` is an opaque string with meaning only to the wallet endpoint which created it. It is recommended that the wallet endpoint include a mechanism to authenticate the contents of auth tokens it issues (for e.g., with an HMAC, or by encryption with a secret symmetric key). This `auth_token` may be used to [`reauthorize`](#reauthorize) future sessions between these dapp and wallet endpoints.
+If the `sign_in_payload` parameter was included but invalid:
```
sdlaver, line 411: This is a lot of errors, and the chances that a client can do something intelligent with these automatically isn't great. WDYT about just having a `INVALID_SIWS_PAYLOAD` error, and using the error message to convey exactly what went wrong?

spec/spec.md
```diff
@@ -508,20 +513,20 @@ get_capabilities
...
-- `supports_sign_and_send_transactions`: `true` if the [`sign_and_send_transactions`](#sign_and_send_transactions) method is supported, otherwise `false`
 - `max_transactions_per_request`: (optional) if present, the max number of transaction payloads which can be signed by a single [`sign_transactions`](#sign_transactions) or [`sign_and_send_transactions`](#sign_and_send_transactions) request. If absent, the implementation doesn't publish a specific limit for this parameter.
 - `max_messages_per_request`: (optional) if present, the max number of transaction payloads which can be signed by a single [`sign_messages`](#sign_messages) request. If absent, the implementation doesn't publish a specific limit for this parameter.
 - `supported_transaction_versions`: the Solana network transaction formats supported by this wallet endpoint. Allowed values are those defined for [`TransactionVersion`](https://solana-labs.github.io/solana-web3.js/modules.html#TransactionVersion) (for e.g., `"legacy"`, `0`, etc).
+- `features`: a list of [feature identifiers](#feature-identifiers) for the optional features supported by this wallet endpoint.
```
sdlaver, line 529: > optional features

Does this mean that the `solana:` namespaced features won't be in this list?

spec/spec.md
```diff
@@ -839,6 +844,80 @@ const ERROR_ATTEST_ORIGIN_ANDROID = -100
 {% include_relative _diagrams/reauthorize_and_sign.plantuml %}
 {% endplantuml %}
 
+### Deprecated Methods
+
+These methods have been deprecated in the current version of the specification, but should still be supported by wallet endpoints for backwards compatibility. These methods may be removed in a future version of the specification.
```
sdlaver, line 849: Rather than repeat the entire message here, WDYT about linking to an archived v1.0 of the spec? It'd also be in GitHub, so no risk of it disappearing on us. It'll help keep this spec concise.

spec/spec.md
Funkatronics: that was my thought, yes. that mandatory features would not be listed here. 

should mandatory features have a specific namespace? or do you think explicitly calling out mandatory vs optional features is sufficient?

spec/spec.md
Funkatronics: yeah I like that

spec/spec.md
Funkatronics: Sort of. Wallet standard does not enforce a payload format, but states that "a wallet may modify the payload before signing, for example to add a message preamble". So wallet standard returns the signature and signed message, which may have been modified by the wallet.

spec/spec.md
Funkatronics: in the response. Will update

spec/spec.md
Funkatronics: yeah I like that. My plan was to maintain a spec1.0.md file and make a spec2.0.md. I may do that in a sperate PR tho (for review purposes I prefer the file diff here on spec.md)

spec/spec.md
Funkatronics: I prefer pervious 😉

spec/spec.md
Funkatronics: Yeah so my thought here is this: the chains in the account metadata are only there to tell a dapp what chains an account can be used on. The dapp would still need to obtain an `auth_token` for the account(s) and chain they want to use. 

Now in practice, this does seem unnecessary/not useful given that a dapp cannot specify which account is authorized, and only one chain is allowed per `auth_token`. So at this time, this field is really just aligning with the wallet standard WalletAccount interface.

spec/spec.md
Funkatronics: oh my mistake, I was referring to the `signMessages` returned payload. the sign in with solana payload is aligned with the [solana wallet standard extension](https://github.com/solana-labs/wallet-standard/blob/master/packages/core/features/src/signIn.ts),  yes.

spec/spec.md
Funkatronics: Some thoughts:

1. Spec version exchanged or reported during key exchange

   - add a version byte to the end of the HELLO_RSP (and possibly the hello request as well)
   - if version byte is missing assume v1.0 
   - dapp/clientlib can decide what to do with the reported version, ie fallback on 1.0 functionality 

2. Because only the RPC interface has changed in this version, we could version the RPC spec and expose it via `get_capabilities`. Lack of a version in the response would indicate MWA 1.0 rpc spec

3. Version RPC methods individually, like Wallet-Standard. Each feature could have an associated version, and `get_capabilities` would expose the supported version of each feature (`["solana:signMessages": "1.0.0", ... ]`)

   - this seems overly complex and prone to loads of complexity for clients. RPC methods should be backwards compatible where ever possible 

4. Do nothing at this time, apart from documentation. v2.0 RPC spec is backwards compatible with v1.0 and association/transport has not been changed. 

I think we should do option 1. its the most robust design for versioning and allows us to make more drastic changes to either the rpc spec of the core transport in the future.

sdlaver: I spent some time thinking about this. I think we need to handshake to select a version, so that we can implement breaking changes. Just querying via `get_capabilities` isn't enough - the client may need to request behavior set `B` that is incompatible with behavior set `A`.

I'd also like to avoid adding a new JSON-RPC message - it adds some complexity, and has the downside of still being able to connect to wallets and then find out the client is not compatible. The only way to prevent that connection in the first place is via the association URI.

Building on your option 1, here's what I propose:
- Any protocol breaking changes MUST rev the version field in the association URI. Currently, this is required to be `v1`. Since MWAv2 is fully backward-compatible, there is no need to rev this value.
- We'll add a new query parameter to the URI, `v=`, which are the major versions that the client supports. This parameter can be repeated. The wallet should select the highest version from this set that it supports.
- If the `v=` query parameter is present, we know that the client is aware of the new versioning scheme. In that case, immediately after the `HELLO_REQ`/`HELLO_RSP` exchange, the wallet should send an unsolicited "session properties" JSON payload to the client. This should be considered part of session establishment, and so is not a JSON-RPC 2.0 response (as there is no corresponding request). This payload should only convey information mandatory to the proper operation of the session (such as the negotiated version). If a client supports legacy connection establishment, failure to receive this session properties indicates that the connection is a legacy connection. A wallet may only send this payload if the `v=` query parameter is present (to avoid sending it to a client that does not understand it). This payload is only valid during session establishment; a client receiving this at any other time should discard it and close the session.
- If a client needs optional features, it must check those using the relevant mechanisms defined in the spec (such as `get_capabilities`. The protocol version is not the appropriate way to convey such information.

In terms of implementation, in MWAv2, we can add support for this scheme (`v=1` query parameter, and updated session establishment protocol when present). Since the rest of the protocol is backward-compatible, we don't need to actually add any new version numbers in the association URI (either in the path, or as query parameters).

WDYT?

spec/spec.md
sdlaver: I agree, I don't think mandatory features need to be included in this list. If they are, you get into a situation where a mandatory feature is missing - in that case, what does a client do? Better if the mandatory features are assumed (based on the negotiated protocol version), since there is nothing a client can realistically do if they are missing.

spec/spec.md
Funkatronics: I like it. I've added a commit capturing all of this.

spec/spec.md
```diff
@@ -270,11 +308,35 @@ Once each endpoint has calculated the ephemeral shared secret, they should proce
...
+
+#### Description
+
+Immediately after sending the `HELLO_RSP` message, the wallet endpoint should follow up with a `SESSION_PROPS` message to convey information mandatory to the proper operation of the session, such as the negotiated version. This message consists of a JSON payload as described above.This payload is only valid during session establishment; a client receiving this at any other time should discard it and close the session.
+
+A wallet may only send this payload if the `v=` query parameter is present in the association URI presented by the client (to avoid sending it to a client that does not understand it). If the connecting dapp endpoint is using a legacy connection (no `v=` parameter present during association), the wallet should not send the `SESSION_PROPS` message to the client. If a client supports legacy connection establishment, failure to receive this session properties message indicates that the connection is a legacy connection.
```
sdlaver, line 333: > A wallet may only send this payload if the `v=` query parameter is present in the association URI presented by the client (to avoid sending it to a client that does not understand it). If the connecting dapp endpoint is using a legacy connection (no `v=` parameter present during association), the wallet should not send the `SESSION_PROPS` message to the client. If a client supports legacy connection establishment, failure to receive this session properties message indicates that the connection is a legacy connection.

I had a hard time interpreting whether this is required or not. I'd recommend restructuring to be of the form:
"If a `v=` query parameter is present in the association URI presented by the dapp endpoint, the wallet endpoint MUST send a `SESSION_PROPS` message to the client immediately after the `HELLO_RSP` message.  If the connecting dapp endpoint runs software for a legacy version of this standard (i.e. no `v=` parameter present during association), the wallet MUST NOT send the `SESSION_PROPS` message to the client, and must assume that the connection is a legacy connection. If the wallet endpoint does not support legacy connections, it should close the connection immediately upon receipt of the `HELLO_REQ` message."

spec/spec.md
sdlaver, line 1: We should update the lead-in materials of the spec to note that this is v2.0.0-DRAFT (or something similar)

sdlaver (approved):

PR #523, opened 2023-08-03
MWA ktx API Final Simplification

creativedrewy:


android/clientlib-ktx/src/main/java/com/solana/mobilewalletadapter/clientlib/MobileWalletAdapter.kt
```diff
@@ -79,11 +97,14 @@ class MobileWalletAdapter(
...
+                                } ?: run {
                                     adapterOperations.authorize(identityUri, iconUri, identityName, rpcCluster)
-                                } else {
-                                    adapterOperations.reauthorize(identityUri, iconUri, identityName, authToken)
                                 }
+
+                                authToken = authResult.authToken
+                                authResult
                             }
                         } else {
```
Funkatronics, line 109: should an error be returned here? or is the thought that calling privileged methods in block will trigger the error if no credentials have been provided?

android/clientlib-ktx/src/main/java/com/solana/mobilewalletadapter/clientlib/MobileWalletAdapter.kt
```diff
@@ -15,6 +15,7 @@ import java.util.concurrent.TimeUnit
 import java.util.concurrent.TimeoutException
 
 class MobileWalletAdapter(
+    connectionIdentity: ConnectionIdentity? = null,
```
Funkatronics, line 18: should this be mandatory (non-optional)? what is the use case for not providing this during construction?

android/clientlib-ktx/src/main/java/com/solana/mobilewalletadapter/clientlib/MobileWalletAdapter.kt
creativedrewy: Returning null is the fallback to keep using it the "old" way of manually doing auth/reauth yourself.

android/clientlib-ktx/src/main/java/com/solana/mobilewalletadapter/clientlib/MobileWalletAdapter.kt
creativedrewy: hmm, using it the old way I think

Funkatronics: In the interest of backwards compatibility, should this be added as the last argument to the constructor?

android/clientlib-ktx/src/main/java/com/solana/mobilewalletadapter/clientlib/MobileWalletAdapter.kt
Funkatronics: discussed offline, no error is returned so that existing implementations will continue to work as is.

android/clientlib-ktx/src/main/java/com/solana/mobilewalletadapter/clientlib/MobileWalletAdapter.kt
creativedrewy: Actually this is a good idea

Funkatronics (approved):

PR #532, opened 2023-08-24
Feature Extensions Template + Ephemeral Signers

Funkatronics:


Funkatronics:
@sdlaver I have added a feature extension template file, and started a draft for the ephemeral signers/accounts feature. Currently it only covers basic ephemeral signer usage like that of [fuse wallet](https://www.npmjs.com/package/@sqds/fuse-wallet). I will sync with Divya next week to nail down this feature exactly.

Funkatronics:
## Ephemeral Wallets, lets goooo!

before we finalize this feature I want to ensure its capturing all our needs.

These are different types of ephemeral accounts/signers that we have identified:
- ephemeral signer
  - a signer that is created during the lifetime of a transaction and used to sign said transaction. this signer has no funds and is never used as a payer. These accounts are expected to be forgotten after the transaction ahs been signed/processed. 
  - because this signer will not hold any funds, it should be invisible to the user. Wallets should generate the keypairs and provide them alongside the users funded accounts without requiring the user to explicitly create or approve them.
- ephemeral payer (aka "burner")
  -  a "burner account" signer that pays for and signs a transaction. The lifetime of this account is determined by the wallet but a dapp can expect that this account will have funds and be available for use for the lifetime of a session/authorization scope. 
  - how these wallets are generated and managed will be a wallet competency. These wallets could be created for a single session and closed when the session is terminated, or could live longer. These are effectively hot wallets that do not hold the users primary assets. Dapps should not make any assumptions about the reuse of these wallets and should therefor not use these public keys for user identity. 

Other **non-ephemeral** account types that will likely be used alongside ephemeral accounts:
- readonly
  - an account that cannot be used to sign transactions. Dapps can use this account to view assets and request `signMessages` to prove ownership of the account. but wallets should reject any requests to sign transactions with this account. This account can be thought of as an identity or login. it should be thought of as the primary account of the user/session

### How do we label and identify these different account types?

#### Initial idea: use feature flags

In MWA 2.0, a list of authorized accounts is returned in the `authorize` response. Each authorized account has a `features` parameter which can be used to specify a list of feature IDs that the account supports. These feature flags could be used to describe the capabilities of these different accounts:
- ephemeral signer
  - `features: [ "solana:signTransaction", "solana:ephemeralSigner" ]`
- ephemeral payer (aka "burner")
  - `features: [ "solana:signMessage"(?), "solana:signTransaction", "solana:payer" (see note) ]` 
  - its hard to distinguish a burner from an ephemeral signer using the existing feature flags. We'd want to add some new flag like `solana:fundTransaction` or `solana:payer`
- readonly
  - `features: [ "solana:signMessage" ]`

Pros: 
- no changes to existing spec (apart from possibly adding a new feature flag for ephemeral signers)
- accounts are identified purely by capability, which makes adding new types easier. its more composable

Cons:
- there is more ambiguity with this approach

#### Alternative: add a new profile param to authorized accounts

Alternatively, we can define a set of account profile IDs and add a new `profile` or `type` (name TBD) param to the authorized account struct. 
- ephemeral signer
  - `profile: "solana:ephemeralSigner"`
- ephemeral payer (aka "burner")
  - `profile: "solana:ephemeralPayer"`
- readonly
  - `profile: "solana:readonly"`

Pros: 
- more explicit, easier to identify exactly which account is which

Cons:
- requires addition of a new parameter. we might want to add this new parameter to the base spec
- new profile types need to added explicitly.

Funkatronics:
will separate template and redo this feature in a new pr

PR #550, opened 2023-09-19
Fix Spec Display Names

Funkatronics:
Currently the 2 spec files have the same title, and get displayed on the web like so:
image

This PR changes their names to:
- Mobile Wallet Adapter 2.0-DRAFT specification 
- Mobile Wallet Adapter 1.0 specification

(DRAFT will be removed when we formally rev the spec to 2.0.0)


creativedrewy (approved):
yay

PR #558, opened 2023-10-02
MWA 2.0 Version Negotiation

Funkatronics:


Funkatronics:
question:

What should a dapp endpoint do if a wallet endpoint does not send the SESSION_PROPS message after the HELLO_RSP message?

scenario:

- dapp is mwa 2.0 compliant, so it sends the `&v=` query parameter during association
- wallet is out of date and using 1.0/legacy specification. it does not know about the `v=` parameter sent from the dapp and does not send the `SESSION_PROPS` message after `HELLO_RSP`
- the dapp endpoint is now stuck waiting for the `SESSION_PROPS` message. it needs to know which protocol version to use for the session before it can allow the client to make requests.

right now im thinking the best way to handle this is either:

1. add a timeout to the client side, if `SESSION_PROPS` is not received within that timeout it falls back on legacy behavior and moves on.
1. Make the session properties a part of the `HELLO_RSP` message so that its not sent as a separate message. This way the dapp can immediately know if the wallet endpoint is legacy when it receives the `HELLO_RSP` message and does not need to wait for a timeout

Funkatronics:
> question:
> 
> What should a dapp endpoint do if a wallet endpoint does not send the SESSION_PROPS message after the HELLO_RSP message?

offline discussion with @sdlaver, we decided to make the session properties part of the HELLO_RSP message instead of a separate message. With a separate message, the dapp endpoint would need to wait for the SESSION_PROPS message before it could move on and begin making requests.

creativedrewy (approved):

Funkatronics:
@creativedrewy my mistake, this is still WIP. I forgot to continue my work after updating the spec with the above discussion.

PR #563, opened 2023-10-06
Align with wallet-standard WalletAccount interface

Funkatronics:
rationale:
- `address` as a base-64 encoded public key is only applicable to Solana. Ethereum addresses use hex strings for example (`0x{hexString}`)
- adding an `address_format` field disambiguates how the address is encoded. we can default to `base64` for backwards compatibly here. 
- `public_key` aligns more with how `address` has been used in 1.0. loosening the required encoding for `address` provides better multichain support.
- `icon` good change for UX IMO. i meant to add this originally and forgot. optional so this is benign


spec/spec.md
```diff
@@ -422,10 +425,13 @@ where:
...
+  - `address_format`: (optional) the format of the `address`. Defaults to `base64` for backwards compatibility with previous versions of this spec. 
+  - `public_key`: a base64-encoded public key for this account
   - `chains`: a list of [chain identifiers](#chain-identifiers) supported by this account. These should be a subset of the chains supported by the wallet.
   - `features`: (optional) a list of [feature identifiers](#feature-identifiers) that represent the features that are supported by this account. These features must be a subset of the features returned by [`get_capabilities`](#get_capabilities). If this parameter is not present the account has access to all available features (both mandatory and optional) supported by the wallet.  
   - `label`: (optional) a human-readable string that describes the account. Wallet endpoints that allow their users to label their accounts may choose to return those labels here to enhance the user experience at the dapp endpoint.
+  - `icon`: (optional) a relative path (from `wallet_uri_base`) to an image asset file of an icon for the account. This may be displayed by the app.
```
Funkatronics, line 434: using a relative uri here like we do with the dapp identity. Wallet-standard uses a data uri with a base64 encoded image (svg, png, etc). should we align with ws and use a data uri?

jordansexton: The icon may not be available at a URL from the base URL (it could be dynamically generated in the wallet, an NFT, etc). You may also not want it to be a URL because this allows tracking. This is why we require data URIs in the WS.

spec/spec.md
```diff
@@ -422,10 +425,13 @@ where:
 
 - `auth_token`: an opaque string representing a unique identifying token issued by the wallet endpoint to the dapp endpoint. The format and contents are an implementation detail of the wallet endpoint. The dapp endpoint can use this on future connections to reauthorize access to [privileged methods](#privileged-methods).
 - `accounts`: one or more value objects that represent the accounts to which this auth token corresponds. These objects hold the following properties:
-  - `address`: a base64-encoded address for this account
+  - `address`: the address for this account. The format of this string will depend on the chain, and is specified by the `addressFormat` field
+  - `address_format`: (optional) the format of the `address`. Defaults to `base64` for backwards compatibility with previous versions of this spec.
```
sdlaver, line 429: Previously, the intent of the `address` field was conflated with the public key (since they are the same for Solana). We selected base64 encoding because this wasn't intended to be a human-visible field. Rather, it was present for preparing transactions. If the dApp wanted to display the address to the user, it was responsible for formatting that value itself. Because of that, this looks like a breaking change to me. The default is base64, so everything will probably work on day 1, but what if a wallet decides to start returning this as, e.g., base58?

I think we should accept that our terminology of `address` was broken and note that. Keep the legacy behavior, where `address` is actually the public key, and add a new field, e.g. `display_address` and `display_address_format`.

WDYT?

Funkatronics: yeah I like that. `address` is effectively a public key and probably should have been called that from the beginning. it made sense for 1.0 being solana specific, we have to eat that now. 

I think we should use `display_*` as you suggest and also add a `public_key` param as an alias for `address`. in the future we could remove `address` in favor of `public_key`

spec/spec.md
```diff
@@ -422,10 +425,14 @@ where:
 
 - `auth_token`: an opaque string representing a unique identifying token issued by the wallet endpoint to the dapp endpoint. The format and contents are an implementation detail of the wallet endpoint. The dapp endpoint can use this on future connections to reauthorize access to [privileged methods](#privileged-methods).
 - `accounts`: one or more value objects that represent the accounts to which this auth token corresponds. These objects hold the following properties:
-  - `address`: a base64-encoded address for this account
+  - `address`: a base64-encoded public key for this account. 
+  - `display_address`: (optional) the address for this account. The format of this string will depend on the chain, and is specified by the `display_address_format` field
+  - `display_address_format`: (optional) the format of the `display_address`.
+  - `public_key`: (optional) a base64-encoded public key for this account. This is an alias for `address` with more accurate terminology. If present, this param should be preferred over `address`.
```
sdlaver, line 431: I'm not sure we should add this as an alias. There's no backward-compat concern at the moment (like with the `chain`/`cluster` parameter to `authorize`). If we do a breaking spec revision in the future, we can rename it then.

sdlaver (changes requested):

spec/spec.md
Funkatronics: removed. (and fixed the `display_*` params)

Funkatronics: We should add some commentary about `address`, in either the description or non-normative section. Could I lean on you @sdlaver for some help with the language here?

sdlaver: How about something like:

> There are 3 parameters in the result that are derived from the unique identity of the account: `address`, `display_address`, and `display_address_format`. `address` contains the public key of the account (for whatever cryptographic system is used as the authority for the account). For example, in the Solana blockchain, this is the Ed25519 public key. `address` should always be encoded with base64, as this parameter is inteded for programmatic consumption rather than human consumption. In contrast, the `display_address` and `display_address_format` are representations of the account appropriate for human consumption. The contents of these two fields are chain-specific; for the Solana blockchain, `display_address_format` will be `base58`, and `display_address` will be the public key for the account, encoded with base58. Other chains may use different encodings and/or different methods for obtaining a human-consumable address.

As an aside, this doesn't address whether the response should include a version of `display_address` that is appropriate for programmatic consumption. Not every blockchain uses this (i.e. Solana does not), so it can always be addressed on a per-chain basis with spec extensions. If we find the need for it is common, we can always add it into future spec versions.

spec/spec.md
sdlaver, line 1: Random aside - I noticed there is an extra backtick at line 452

spec/spec.md
```diff
@@ -461,13 +467,15 @@ The `chain` parameter allows the dapp endpoint to select a specific chain with w
...
 
+There are 3 parameters in the result that are derived from the unique identity of the account: address, display_address, and display_address_format. address contains the public key of the account (for whatever cryptographic system is used as the authority for the account). For example, in the Solana blockchain, this is the Ed25519 public key. address should always be encoded with base64, as this parameter is inteded for programmatic consumption rather than human consumption. In contrast, the display_address and display_address_format are representations of the account appropriate for human consumption. The contents of these two fields are chain-specific; for the Solana blockchain, display_address_format will be base58, and display_address will be the public key for the account, encoded with base58. Other chains may use different encodings and/or different methods for obtaining a human-consumable address.
```
sdlaver, line 470: Need a few backticks here to add context to when a word (i.e. `address`) is referring to a specific parameter

sdlaver (approved):

spec/spec.md
Funkatronics: on it!

PR #569, opened 2023-10-11
Properly implement 2.0 authorize method in MobileWalletAdapterClient

Funkatronics:


Funkatronics:
@creativedrewy this will unblock you but I am not loving this solution. 

The `MobileWalletAdapterClient` in `clientlib` does not know about the session properties and because both chain and cluster are strings, there is no way to disambiguate the call. 

some other ideas:
- make a `MobileWalletAdapterClient.authorizeV2` method that takes chain, and revert the `MobileWalletAdapterClient.authorize` method to use cluster so clients know exactly what they are calling. 
- make an `Identifier` or `ChainIdentifier` object instead of using strings for chains. This would help disambiguate a legacy `authorize` method from a new one that takes the chain object

Funkatronics:
Summary of discussions with @creativedrewy 

These changes will work, tho may not be ideal. 

`clientlib.MoibleWalletAdapterClient` exposes the new 2.0 `authorize` functionality, while preserving the legacy behavior. The downside of this is that a client might inappropriately call the 2.0 authorize on a 1.0 wallet endpoint. The client will need to be aware of the reported session properties and do the right thing. Ideally, we'd like the scenario/session to only expose the new 2.0 functionality when a 2.0 session has been established, and only expose 1.0/legacy functionality otherwise.

Funkatronics:
tests pass locally, flakiness on ci. merging @creativedrewy

PR #581, opened 2023-10-19
[WIP] Deprecate signTransactions in libs

Funkatronics:


Funkatronics:
@creativedrewy this might break the example ktx app

creativedrewy (approved):
I actually think KTX stuff should be good!

Funkatronics:
ty drewy. gonna merge this pr, tho we still need to think about the implications of deprecating signTransaction and backwards compat.

PR #584, opened 2023-10-24
[WIP] Sign in with Solana

Funkatronics:


Funkatronics:
Review notes:

This pr is quite large, but can be broken down by lib
- `common`
  - contains the core SIWS payload struct and message+json parsing 
- `walletlib`
  - added the `sign_in_payload` to the `authorize` request and forwards this to the wallet for handling (if present)
  - added the `sign_in_result` to the authorization result so wallets can pass the signed payload back to the dapp along with the authorize request
- `client lib`
  - added the (optional) `sign_in_payload` to the `authorize` request input and returns the `sign_in_result` from the wallet

- `fakedapp`
  - added a "Sign in with Solana" button that does `authorize` with a `sign_in_payload` 
  - if the wallet fails to return the `sign_in_result`, it falls back on `sign_message` immediately after the authorize request is successfully completed 
    - this could be handled in `clientlib` but it required significant refactor so left it here for now. `clientlib-ktx` could easily handle this for the consumer however (cc @creativedrewy).
- `fakewallet`
  - adds a new authorization fragment that specifically handles an `authorize` request with a `sign_in_payload` 
    - There is also a "Simulate Sign In Not Supported" button on this UI that will complete the authorize request and ignore the signin payload, to simulate a legacy wallet, or a wallet that does not the solana:signInWithSolana feature

android/clientlib/src/main/java/com/solana/mobilewalletadapter/clientlib/protocol/MobileWalletAdapterClient.java
```diff
@@ -145,13 +146,17 @@ public static class AuthorizationResult {
         public final Uri walletUriBase;
         @NonNull @Size(min = 1)
         public final AuthorizedAccount[] accounts;
+        @Nullable
+        public final SignInResult signInResult;
```
creativedrewy, line 150: The ergonomics of this object feels...odd, given its overlap with some of the values in the `AuthorizedAccount` object. But I don't really think there is a better option than this. I think you were forced to bolt this on.

creativedrewy (approved):
A lotta plumbing for...something useful?

android/clientlib/src/main/java/com/solana/mobilewalletadapter/clientlib/protocol/MobileWalletAdapterClient.java
Funkatronics: yeah im not thrilled with it. I could subclass the authorize objects but that would add complexity elsewhere so this felt the least bad (lol).

PR #604, opened 2023-11-13
Remove RxJava Lib + Example App

Funkatronics:
remove `clientlib-rxjava` and its corresponding example app. This library is not used and therefore not worth maintaining.


creativedrewy (approved):
Bye bye

PR #619, opened 2023-11-28
MWA 2.0 Web: Association

Funkatronics:


js/packages/mobile-wallet-adapter-protocol/src/transact.ts
```diff
@@ -171,7 +173,18 @@ export async function transact<TReturn>(
                         state.associationPublicKey,
                         state.ecdhPrivateKey,
                     );
-                    state = { __type: 'connected', sharedSecret };
+                    const sessionPropertiesBuffer = responseBuffer.slice(ENCODED_PUBLIC_KEY_LENGTH_BYTES);
```
Michaelsulistio, line 176: Testing my understanding here:

In v1 protocol (MWA 2.0), the wallet will include a `session_properties` in their `responseBuffer`. 

It will now be `responseBuffer = public_key + session_properties` (all encoded).

js/packages/mobile-wallet-adapter-protocol/src/parseHelloRsp.ts
```diff
@@ -14,7 +16,7 @@ export default async function parseHelloRsp(
         crypto.subtle.exportKey('raw', associationPublicKey),
         crypto.subtle.importKey(
             'raw',
-            payloadBuffer,
+            payloadBuffer.slice(0, ENCODED_PUBLIC_KEY_LENGTH_BYTES),
```
Michaelsulistio, line 19: This change is necessary because the response buffer now contains more than just the public key. It now includes also `session_properties`

js/packages/mobile-wallet-adapter-protocol/src/transact.ts
```diff
@@ -171,7 +173,18 @@ export async function transact<TReturn>(
...
+                    const sessionProperties = sessionPropertiesBuffer.byteLength !== 0 
+                        ? await (async () => {
+                            const sequenceNumberVector = sessionPropertiesBuffer.slice(0, SEQUENCE_NUMBER_BYTES);
```
Michaelsulistio, line 179: Testing my understanding:

There's now a `sequenceNumber` also contained within the `sessionProperties` buffer

Michaelsulistio (approved):

js/packages/mobile-wallet-adapter-protocol/src/parseHelloRsp.ts
Funkatronics: yes, well it might include the session properties. Subtle crypto was throwing an error if I did not slice off the extra bytes. 

if the session properties are not present (ie when interacting with a legacy wallet), this is still safe.

js/packages/mobile-wallet-adapter-protocol/src/transact.ts
Funkatronics: the session properties are encrypted, in the same way we encrypt ever other payload that is sent over the websocket following the key exchange. 

more details about the encrypted message wrapping in the spec [here](https://solana-mobile.github.io/mobile-wallet-adapter/spec/spec.html#encrypted-message-wrapping).

PR #621, opened 2023-11-29
feat: constrain the connection interface of `sendTransaction()` to only that which is required

steveluscher:
Related to the soon-to-be-released https://github.com/solana-labs/wallet-adapter/pull/848.

### DIFF
--- a/js/packages/wallet-adapter-mobile/src/adapter.ts
+++ b/js/packages/wallet-adapter-mobile/src/adapter.ts
@@ -1,5 +1,6 @@
 import {
     BaseMessageSignerWalletAdapter,
+    ConnectionContext,
     WalletConnectionError,
     WalletDisconnectedError,
     WalletName,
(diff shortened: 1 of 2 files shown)


js/packages/wallet-adapter-mobile/src/adapter.ts
```diff
@@ -394,15 +394,11 @@ export class SolanaMobileWalletAdapter extends BaseMessageSignerWalletAdapter {
...
-                        if (isVersionedTransaction(signedTransaction)) {
-                            return await connection.sendTransaction(signedTransaction);
-                        } else {
-                            const serializedTransaction = signedTransaction.serialize();
-                            return await connection.sendRawTransaction(serializedTransaction, {
-                                ...options,
-                                preflightCommitment: getTargetCommitment(),
-                            });
-                        }
+                        const serializedTransaction = signedTransaction.serialize();
```
steveluscher, line 397: Both `LegacyTransaction` and `VersionedTransaction` implement `serialize()` so there's no reason to disambiguate.

Funkatronics (approved):

PR #623, opened 2023-11-30
MWA 2.0 Web: 2.0 API

Funkatronics:


js/packages/mobile-wallet-adapter-protocol/src/types.ts
```diff
@@ -54,16 +60,33 @@ type Base64EncodedSignedTransaction = string;
...
+     */
     authorize(params: { cluster: Cluster; identity: AppIdentity }): Promise<AuthorizationResult>;
+
+    authorize(params: { 
+        identity: AppIdentity;
+        chain?: Chain;
+        features?: IdentifierArray; 
+        addresses?: string[]; 
+        auth_token?: AuthToken; 
+        sign_in_payload?: SolanaSignInInput;
```
Michaelsulistio, line 88: What happens when a v1 dApp tries to authorize with `sign_in_payload` using a legacy wallet? IIUC, the backwards compatibility isn't handled and the wallet will try a normal authorize.

How can the dApp check if the wallet supports SIWS?

Michaelsulistio: answered offline:
> it would need to call getCapabilities, just as normal/native

Michaelsulistio (approved):

js/packages/mobile-wallet-adapter-protocol/src/types.ts
Funkatronics: a dapp can either call `getCapabilities` before hand, or it can simply call `authorize` with a sign in payload and check for a sign in result in the response. 

In a separate PR I will address SIWS more completely.

PR #633, opened 2023-12-08
MWA JS: Sign in with Solana

Funkatronics:


js/packages/mobile-wallet-adapter-protocol/src/types.ts
```diff
@@ -85,7 +85,7 @@ export interface AuthorizeAPI {
         features?: IdentifierArray; 
         addresses?: string[]; 
         auth_token?: AuthToken; 
-        sign_in_payload?: SolanaSignInInput;
+        sign_in_payload?: Pick<SignInPayload, 'domain' | 'uri'> & Partial<SignInPayload>;
```
Michaelsulistio, line 88: IIUC, the `SignInPayload` type is used 'internally' by the SDK, but this utility type will be used by exposed/consumed by devs.

For type readability and usability by devs, I'd suggest moving this into a public type like:

```
export type SolanaSignInPayload = Pick<SignInPayload, 'domain' | 'uri'> & Partial<SignInPayload>;
```

Similar to how `Chain`, `AuthToken`, etc are defined. Type name is more readable and more importantly, devs can import and use the type.

js/packages/mobile-wallet-adapter-protocol/src/createSIWSMessage.ts
```diff
@@ -0,0 +1,22 @@
+import { Buffer } from 'buffer'
+import { SignInPayload } from './types';
+import { SolanaSignInInputWithRequiredFields, createSignInMessageText } from '@solana/wallet-standard-util';
+
+export function createDefaultSolanaSignInParams(): Pick<SignInPayload, 'version' | 'chainId' | 'nonce' | 'issuedAt' > {
+    return {
+        version: '1',
+        chainId: '1',
+        nonce: Buffer.from(
+            crypto.getRandomValues(new Uint8Array(8))
```
Michaelsulistio, line 10: Sanity check here: Where does the `crypto` module come from? Does this work in react native app's that don't import some `crypto` polfyill?

Michaelsulistio (approved):

js/packages/mobile-wallet-adapter-protocol/src/types.ts
```diff
@@ -154,3 +154,25 @@ export interface MobileWallet
...
+    requestId?: string;
+    resources?: string[];
+}> & SolanaSignInInput;
+
+export type SignInResult = Readonly<{
+    publicKey: Base64EncodedAddress;
+    signed_message: Base64EncodedSignedMessage;
+    signature: Base64EncodedAddress;
+    signature_type?: string;
+}>; // | SolanaSignInOutput;
```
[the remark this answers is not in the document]
Funkatronics: leaving it for now until SIWS is all wrapped up

js/packages/mobile-wallet-adapter-protocol/src/createSIWSMessage.ts
Funkatronics: we chatted about this on slack but posting her for viz:

this does require that the react native app uses the `react native get random values` polyfill. this was already a requirement so nothing has changed there.

js/packages/mobile-wallet-adapter-protocol/src/types.ts
Funkatronics: SignInPayload can/will be used publicy as well. 

but yeah I can move this to a type like [`SolanaSignInInputWithRequiredFields`](https://github.com/solana-labs/wallet-standard/blob/2a1d55c9a57170f8ccecbb2f81f070442e7bd670/packages/core/util/src/signIn.ts#L56C48-L56C48)

PR #634, opened 2023-12-12
fix bug with null statement in siws payload

Funkatronics:
fixes a small bug in the SIWS impl where a null `statement` would show up as the text "null" in the sign in message. this change ensures that a null `statement` results in no statement text being added in the output message


Funkatronics:
superseded by #636

PR #635, opened 2023-12-13
wallet-adapter-mobile siws support

Funkatronics:


Funkatronics:
Note: testing this PR relies building fakewallet from #636

js/packages/mobile-wallet-adapter-protocol/src/createMobileWalletProxy.ts
```diff
@@ -31,8 +31,8 @@ export default function createMobileWalletProxy<
                     const { method, params } = handleMobileWalletRequest(p, inputParams, protocolVersion);
                     const result = await protocolRequestHandler(method, params) as Awaited<ReturnType<MobileWallet[TMethodName]>>;
                     // if the request tried to sign in but the wallet did not return a sign in result, fallback on message signing
-                    if (method === 'authorize' && params?.hasOwnProperty('sign_in_payload') && !result.hasOwnProperty('sign_in_result')) {
-                        (result as any)['sign_in_result'] = signInFallback(
+                    if (method === 'authorize' && (params as any).sign_in_payload && !(result as any).sign_in_result) {
```
Michaelsulistio, line 34: why not `params?.sign_in_payload` and `result?.sign_in_result`?

js/packages/wallet-adapter-mobile/src/adapter.ts
```diff
@@ -233,6 +221,33 @@ export class SolanaMobileWalletAdapter extends BaseMessageSignerWalletAdapter {
...
+    async performAuthorization(signInPayload?: SignInPayload): Promise<AuthorizationResult> {
+        try {
+            const cachedAuthorizationResult = await this._authorizationResultCache.get();
+            if (cachedAuthorizationResult) {
+                // TODO: Evaluate whether there's any threat to not `awaiting` this expression
+                this.handleAuthorizationResult(cachedAuthorizationResult);
+                return cachedAuthorizationResult;
+            }
+            return await this.transact(async (wallet) => {
+                const authorizationResult = await wallet.authorize({
```
Michaelsulistio, line 233: What happens if a SIWS request occurs but the mobile wallet doesn't support SIWS? 

Would the `wallet.authorize` call throw an error (good) or simply ignore the signInPayload and use normal `authorize`?

js/packages/mobile-wallet-adapter-protocol/src/createMobileWalletProxy.ts
```diff
@@ -89,13 +89,6 @@ function handleMobileWalletRequest<TMethodName extends keyof MobileWallet>(
                     case 'mainnet-beta': { chain = 'solana:mainnet'; break; }
                 }
                 (params as Parameters<MobileWallet['authorize']>[0]).chain = chain;
-
-                if (signInPayload) {
```
Michaelsulistio, line 93: Checking understanding:

This is now removed because `params` will now just contain a `SignInPayload` type that doesn't need any conversion.

Michaelsulistio (approved):

js/packages/mobile-wallet-adapter-protocol/src/createMobileWalletProxy.ts
Funkatronics: because the type of param is generic (boxed to ReurnType<MobileWallet>), so you cannot do this.

PR #636, opened 2023-12-13
Align SIWS implementation with Phantom spec

Funkatronics:
Aligns the native java SIWS implementation with the Phantom SIWS spec: https://github.com/phantom/sign-in-with-solana


Michaelsulistio (approved):
LG!

PR #647, opened 2023-12-20
Clientlib Ktx SIWS

Funkatronics:
Adds support for SIWS to the ktx lib, and exposes the functionality in the ktx sample app.


examples/example-clientlib-ktx-app/app/src/main/java/com/solanamobile/ktxclientsample/viewmodel/SampleViewModel.kt
```diff
@@ -62,116 +69,114 @@ class SampleViewModel @Inject constructor(
...
-                }
-                is TransactionResult.NoWalletFound -> {
-                    _state.value.copy(
-                        walletFound = false
-                    ).updateViewState()
+    fun signIn(sender: ActivityResultSender) {
+        viewModelScope.launch {
+            connect(sender) {}
+            // Note: should check the signature here of the signInResult to verify it matches the
+            // account and expected signed message. This is left as an exercise for the reader.
```
Michaelsulistio, line 114: 😂

If trivial we should eventually have a snippet of this. Although true SIWS verification should happen on the server side anyways.

Michaelsulistio (approved):

PR #687, opened 2024-01-25
fix legacy auth call

Funkatronics:


Funkatronics:
@ankur2136 this will fix your problem

ankur2136 (approved):

PR #749, opened 2024-03-13
SPIKE: Wallet Automated Testing

Funkatronics:
Experimental testing script to test fakedapp against Phantom 

setup:
- setup a phone or emulator to run tests from Android Studio
- manually install Phantom apk on device
- ensure phantom is setup with a wallet
- ensure phantom is set to test/dev mode with the solana testnet cluster selected

to run:
run the tests in the `MainActivityTestsPhantom` file in fakedapps test package

TODOs:
- ~~need to investigate why Phantom is returning invalid signatures with this script~~
- ~~investigate mainnet/testnet discrepancy mention above~~
- rethink the approach here. Should we be using fakedapp? it might be better to simulate/mock the client side like we do in the fakewallet tests

### DIFF
--- a/.github/workflows/android.yml
+++ b/.github/workflows/android.yml
@@ -147,8 +147,8 @@ jobs:
         run: |
           cd android
           adb shell am broadcast -a android.intent.action.CLOSE_SYSTEM_DIALOGS
-          ./gradlew :fakewallet:connectedV1DebugAndroidTest
-          ./gradlew :fakewallet:installV1Debug :fakedapp:connectedDebugAndroidTest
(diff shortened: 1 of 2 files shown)


Funkatronics:
NOTE: CI should fail if it tries to run these tests. they currently fail due to an invalid signature returned from Phantom (using latest version from dapp store)

Funkatronics:
tests here should start working once #752 is merged

Funkatronics:
tests are passing locally with Phantom setup on an emulator. CI needs to be adjusted to only run the fakewallet tests and ignore these phantoms tests for now.

Funkatronics:
not planned, closing

PR #750, opened 2024-03-14
Fix: fakedapp legacy cluster param

Funkatronics:
the kt and react/web libraries handle backwards compatibility automatically, while the core java SDK does not (significant refactor was needed to make the java client session aware so we opted to handle it in the wrapper libs that most use). This PR fixes an issue with fakedapp not handling the backwards compat with legacy wallets (`chain` vs `cluster` param).

### DIFF
--- a/android/fakedapp/src/main/java/com/solana/mobilewalletadapter/fakedapp/MainViewModel.kt
+++ b/android/fakedapp/src/main/java/com/solana/mobilewalletadapter/fakedapp/MainViewModel.kt
@@ -460,17 +460,7 @@ class MainViewModel(application: Application) : AndroidViewModel(application) {
         signInPayload: SignInWithSolana.Payload? = null
     ): MobileWalletAdapterClient.AuthorizationResult {
         val result = try {
-            if (uiState.value.sessionProtocolVersion == ProtocolVersion.V1) {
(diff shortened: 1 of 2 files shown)


android/fakedapp/src/main/java/com/solana/mobilewalletadapter/fakedapp/MainViewModel.kt
```diff
@@ -456,11 +456,21 @@ class MainViewModel(application: Application) : AndroidViewModel(application) {
     private suspend fun doAuthorize(
         client: MobileWalletAdapterUseCase.Client,
         identity: MobileWalletAdapterUseCase.DappIdentity,
-        cluster: String?,
+        chain: String?,
         signInPayload: SignInWithSolana.Payload? = null
     ): MobileWalletAdapterClient.AuthorizationResult {
         val result = try {
-            client.authorize(identity, cluster, signInPayload)
+            if (uiState.value.sessionProtocolVersion == ProtocolVersion.V1) {
```
Funkatronics, line 463: clientlib-kt and the js libs handle this internally, but fakedapp needs to do it manually.

Michaelsulistio (approved):

PR #782, opened 2024-04-10
DRAFT: Account Profiles Specification

Funkatronics:


spec/spec.md
```diff
@@ -416,6 +427,7 @@ where:
             "display_address_format": "<display_address_format>",
             “label”: “<label>”, 
             "icon": "<icon>",
+            "profile": "<account_profile_id>",
```
Funkatronics, line 430: caveat here: adding this parameter to the account struct deviates from the wallet-standard [Wallet Account API](). To support this on the web we will need to map this profile id into the `features` parameter. 

Alternatively, we could do the opposite. We could drop this new parameter in the spec, and instead use the existing `features` parameter. In the mobile SDKs, we could abstract this into an explicit `profile` parameter.

Funkatronics: @sdlaver

sdlaver: I'd put it into the `features` parameter.

spec/spec.md
```diff
@@ -126,6 +126,17 @@ These features are mandatory and must be implemented by wallet endpoints. Dapp e
...
+#### Account Profiles
+
+Account profiles are identifiers that are used to decribe an authorized account's capabilities. These are defined as feature identifiers for compatibiblity with the wallet-standard account features API.
+
+- `solana:readOnly`
+   - A read-only account. These accounts should not be used as writeable accounts in transaction. These acounts may contain assets, but dapp endpoints should only expect to read from these accounts and prove ownership via a message sign. If a dapp requests a transaction signature from a readonly account, the wallet should reject the transaction. Read-only accounts are limited to the following features: [`solana:signMessages`](#sign_messages), `solana:signInWithSolana`.
+- `solana:readWrite`
+   - An account that can be used as a writeable acount in transactions. Read-write accounts support, at a minimum, the following features: [`solana:signMessages`](#sign_messages), [`solana:signAndSendTransaction`](#sign_and_send_transactions).
+- `solana:payer`
+   - An account that will be used to pay for transactions. Payer accounts support, at a minimum, the following features: [`solana:signAndSendTransaction`](#sign_and_send_transactions).
```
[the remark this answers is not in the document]
Funkatronics: The latter.

spec/spec.md
```diff
@@ -139,6 +139,19 @@ These features are deprecated, but can be supported by wallet endpoints to maint
 
 - [`solana:signTransactions`](#sign_transactions)
 
+## Account Profiles
```
Funkatronics, line 142: With this approach, I think it makes more sense to call these "Account Labels"

spec/spec.md
```diff
@@ -139,6 +139,19 @@ These features are deprecated, but can be supported by wallet endpoints to maint
...
+
+Account profiles are labels that are used to decribe an authorized account's capabilities and intended use. Each authorized account that is returned in [`authorize`](#authorize) request response can include a list of these profiles. The dapp endpoint can use these labels in combination with the account's list of [`features`](#feature-identifiers) to determine how each acount should be used. 
+
+- `primary`
+  - The users primary account. This account will be used for sign in, and is expected to have the users primary assets. 
+- `feePayer`
+  - Fee payer accounts, when present, should be used to pay for transaction fees. This profile is incompatible with the `immutable` profile, a fee payer account cannot be `immutable`. 
+- `mutable`
```
sdlaver, line 150: Do we need both `mutable` and `immutable`, or does the absence of one imply the other?

spec/spec.md
```diff
@@ -139,6 +139,19 @@ These features are deprecated, but can be supported by wallet endpoints to maint
 
 - [`solana:signTransactions`](#sign_transactions)
 
+## Account Profiles
+
+Account profiles are labels that are used to decribe an authorized account's capabilities and intended use. Each authorized account that is returned in [`authorize`](#authorize) request response can include a list of these profiles. The dapp endpoint can use these labels in combination with the account's list of [`features`](#feature-identifiers) to determine how each acount should be used. 
+
+- `primary`
```
sdlaver, line 146: Does `primary` imply mutability, or would you always expect this to be accompanied by a mutability label?

spec/spec.md
sdlaver: I think account labels makes sense, adding context of when an account should be used. One potential one I see missing from this list is `identity` - an account that contains a users preferred identity credentials.

spec/spec.md
Funkatronics: My thought is, primary will not imply mutability.

spec/spec.md
Funkatronics: I could go either way, I am not certain what is best. both have drawbacks I think

A) With both `mutable` and `immutable`, its more explicit. tho it also adds the incompatibilities (`mutable` and `feepayer` cannot also be `immutable`)

B) With only `mutable`, the absence would imply immutability. `feePayer` would imply `mutable` as well which makes sense. `primary` would not imply mutable, it would need to be specified. 

C) With only `immutable`, the absence would imply mutability. `feePayer` would be incompatible with `immutable` as it would not make sense to have an immutable fee payer. 

I am leaning towards option B here, only having `mutable`. `primary` (and `identity`) accounts imply immutability if not specified. `feePayer` and `mutable` would be allowed but redundant.

spec/spec.md
```diff
@@ -139,6 +139,21 @@ These features are deprecated, but can be supported by wallet endpoints to maint
 
 - [`solana:signTransactions`](#sign_transactions)
 
+## Account Labels
+
+Account labels are used to decribe an authorized account's capabilities and intended use. Each authorized account that is returned in [`authorize`](#authorize) request response can include a list of these labels. The dapp endpoint can use these labels in combination with the account's list of [`features`](#feature-identifiers) to determine how each acount should be used.
```
sdlaver, line 144: I'd add some commentary here that wallets are encouraged to add labels, and that, when any labels are present, the wallet must fully label all returned accounts.

spec/spec.md
sdlaver: I can think of one benefit for option C - if a wallet labels nothing. The default (and backward-compatible) behavior for an unlabeled account would be mutable.

spec/spec.md
```diff
@@ -442,6 +458,7 @@ where:
   - `features`: (optional) a list of [feature identifiers](#feature-identifiers) that represent the features that are supported by this account. These features must be a subset of the features returned by [`get_capabilities`](#get_capabilities). If this parameter is not present the account has access to all available features (both mandatory and optional) supported by the wallet.  
   - `label`: (optional) a human-readable string that describes the account. Wallet endpoints that allow their users to label their accounts may choose to return those labels here to enhance the user experience at the dapp endpoint.
   - `icon`: (optional) a data URI containing a base64-encoded SVG, WebP, PNG, or GIF image of an icon for the account. This may be displayed by the app.
+  - `labels`: a list of [account labels](#account-labels) for this account.
```
sdlaver, line 461: Is this mandatory or optional?

Funkatronics: should be optional, for backwards compatibility. will update

spec/spec.md
Funkatronics: yeah thats a good point. gonna mull on this a bit more.

PR #840, opened 2024-05-22
Bump rexml version (security alert #71)

Funkatronics:


Funkatronics:
built app and tested locally

PR #903, opened 2024-07-18
Walletlib React Native v1.3.0

Michaelsulistio:
# Walletlib RN Changes

## useMobileWalletAdapterSession 
The change here is motivated by the problematic default `Linking.getInitialUrl` for getting hte `associationUri`. `Linking.getInitialUrl` can be `null` in certain cases, like with Expo Router, and it's faulty to expect that it will always be populated correctly.

- Removed default `associcationUri` using `Linking.getInitialUrl` due to Expo Router incompatibility
- Added an `associationUri` parameter to `useMobileWalletAdapterSession` **(breaking)**

## New API methods
The goal is to expose the API outside of a React hook. This gives users the option to create their own hook to manage a session, decouple event listening and session initialization, and overall flexibility.

- Added an `initializeMWAEventListener` method that starts and returns a bridged listener for MWA events.
- Added an `initializeMobileWalletAdapterSession` method that starts an MWA scenario for the given `associationUri`.

### DIFF
--- a/examples/example-react-native-wallet/screens/MobileWalletAdapterEntrypointBottomSheet.tsx
+++ b/examples/example-react-native-wallet/screens/MobileWalletAdapterEntrypointBottomSheet.tsx
@@ -24,6 +24,8 @@ import {
   DeauthorizeDappResponse,
   MWARequestFailReason,
   getCallingPackage,
+  initializeMWAEventListener,
+  initializeMobileWalletAdapterSession,
 } from '@solana-mobile/mobile-wallet-adapter-walletlib';
 
 import AuthenticationScreen from '../bottomsheets/AuthenticationScreen';
@@ -37,7 +39,10 @@ import {
   VerificationFailed,
   VerificationSucceeded,
 } from '../utils/ClientTrustUseCase';
-import { SolanaSignInWithSolana, SolanaSignTransactions } from '@solana-mobile/mobile-wallet-adapter-protocol';
+import {
+  SolanaSignInWithSolana,
+  SolanaSignTransactions,
+} from '@solana-mobile/mobile-wallet-adapter-protocol';
 import SignInScreen from '../bottomsheets/SignInScreen';
 
 type SignPayloadsRequest = SignTransactionsRequest | SignMessagesRequest;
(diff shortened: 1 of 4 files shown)


examples/example-react-native-wallet/screens/MobileWalletAdapterEntrypointBottomSheet.tsx
```diff
@@ -99,6 +107,30 @@ export default function MobileWalletAdapterEntrypointBottomSheet() {
     setCurEvent(sessionEvent);
   }, []);
 
+  // Initialize the MWA by:
+  //  1. Begin listening for MWA events
+  //  2. Starting the MWA session
+  useEffect(() => {
```
Michaelsulistio, line 113: Refactored to use the non-hook API.

examples/example-react-native-wallet/screens/MobileWalletAdapterEntrypointBottomSheet.tsx
```diff
@@ -184,25 +217,6 @@ export default function MobileWalletAdapterEntrypointBottomSheet() {
     }
   }, [wallet, curRequest, endWalletSession, clientTrustUseCase]);
 
-  // Start an MWA session
-
-  useEffect(() => {
```
Michaelsulistio, line 189: Accidentally copy/pasted a redundant useEffect here

examples/example-react-native-wallet/screens/MobileWalletAdapterEntrypointBottomSheet.tsx
```diff
@@ -184,25 +217,6 @@ export default function MobileWalletAdapterEntrypointBottomSheet() {
...
-    if (!curEvent) {
-      return;
-    }
-
-    if (curEvent.__type === MWASessionEventType.SessionTerminatedEvent) {
-      endWalletSession();
-    }
-  }, [curEvent, endWalletSession]);
-
-  useMobileWalletAdapterSession(
```
Michaelsulistio, line 199: Removed this and refactored to use decoupled API

Funkatronics (approved):

PR #906, opened 2024-07-26
Return a wallet icon in authorize result

Funkatronics:


android/clientlib/src/main/java/com/solana/mobilewalletadapter/clientlib/protocol/MobileWalletAdapterClient.java
```diff
@@ -331,6 +335,14 @@ protected AuthorizationResult processResult(@Nullable Object o)
                         walletUriBaseStr + "'; expected a 'https' URI");
             }
 
+            final Uri walletIcon;
+            try {
+                final String walletIconStr = jo.getString(ProtocolContract.RESULT_WALLET_ICON);
+                walletIcon = Uri.parse(walletIconStr);
+            } catch (JSONException e) {
+                throw new JsonRpc20InvalidResponseException("expected a wallet_icon");
```
Michaelsulistio, line 343: Will we run into this case if the wallet endpoint is on an older version, without the DefaultWalletIconProvider changes?

Funkatronics: oh yes good call

Michaelsulistio (approved):

android/walletlib/src/main/java/com/solana/mobilewalletadapter/walletlib/scenario/AuthorizedAccount.java
```diff
@@ -21,6 +21,9 @@ public class AuthorizedAccount {
     @Nullable
     public final String accountLabel;
     @Nullable
+    public final Uri accountIcon;
+    @Deprecated
```
sdlaver, line 25: Can we add a deprecation description here, so that consumers know to use `accountIcon` instead?

android/walletlib/src/main/java/com/solana/mobilewalletadapter/walletlib/protocol/MobileWalletAdapterServer.java
```diff
@@ -256,6 +256,7 @@ private void onAuthorizationComplete(@NonNull NotifyOnCompleteFuture<Authorizati
                 }
                 o.put(ProtocolContract.RESULT_ACCOUNTS, accounts);
                 o.put(ProtocolContract.RESULT_WALLET_URI_BASE, result.walletUriBase); // OK if null
+                o.put(ProtocolContract.RESULT_WALLET_ICON, result.walletIcon);
```
sdlaver, line 259: We should have a check (somewhere, not necessarily here) that the scheme of this URI is `data:`

sdlaver (changes requested):

android/walletlib/src/main/java/com/solana/mobilewalletadapter/walletlib/scenario/AuthorizedAccount.java
Funkatronics: AFAIK, there is no way to add a deprecation description with the @Deprecated annotation used here. the only fields it accepts are `since` and `forRemoval`. I could add a JavaDoc message instead?

android/walletlib/src/main/java/com/solana/mobilewalletadapter/walletlib/protocol/MobileWalletAdapterServer.java
Funkatronics: I added 2 checks: 1 in the `LocalScenario` constructor to check the uri from the icon provider, and another in the `AuthorizationResult` constructor to check the icon uri parameter

android/walletlib/src/main/java/com/solana/mobilewalletadapter/walletlib/scenario/AuthorizedAccount.java
Funkatronics: I added a JavaDoc comment with the deprecation message. this notifies consumers to use `accountIcon` instead in the IDE tooltip

sdlaver (approved):

PR #917, opened 2024-08-05
 SIWS: disambiguate address encoding

Funkatronics:


android/common/src/main/java/com/solana/mobilewalletadapter/common/signin/SignInWithSolana.java
```diff
@@ -158,15 +203,15 @@ public String prepareMessage() {
             if (domain == null) {
                 throw new IllegalStateException("cannot prepare sign in message, no domain provided");
             }
-            if (address == null) {
+            if (addressRaw == null) {
```
Michaelsulistio, line 206: We're not concerned about backwards compatibility with the deprecated `address` right?

Michaelsulistio (approved):

android/common/src/main/java/com/solana/mobilewalletadapter/common/signin/SignInWithSolana.java
Funkatronics: this is backwards compatible. the existing constructors are still here, just marked for deprecation.

Funkatronics: see here: https://github.com/solana-mobile/mobile-wallet-adapter/blob/df9219b6e87ab96c896f07743457186110c2f56b/android/common/src/main/java/com/solana/mobilewalletadapter/common/signin/SignInWithSolana.java#L176

PR #954, opened 2024-09-09
Mwa wallet standard

Funkatronics:


js/packages/wallet-standard-mobile/package.json
```diff
@@ -0,0 +1,50 @@
+{
+    "name": "@solana-mobile/wallet-standard-mobile",
+    "description": "A wallet-standard wallet for mobile wallet apps that conform to the Solana Mobile Wallet Adapter protocol",
+    "version": "0.1.0",
```
Michaelsulistio, line 4: Do we want to  begin using the `-betaX` semantic, if we're going to be pushing lots of breaking changes to this?

Michaelsulistio (approved):

PR #979, opened 2024-10-04
Fix expired reauthorizations

Funkatronics:


android/walletlib/src/main/java/com/solana/mobilewalletadapter/walletlib/authorization/AuthRepositoryImpl.java
J909, line 1: Really just a naming nit in the context of Room, because "migration" is used on the DB when schemas is updated. But I understand the parentId has to be retained on reauthorization.

J909 (approved):

android/walletlib/src/main/java/com/solana/mobilewalletadapter/walletlib/authorization/AuthRepositoryImpl.java
Funkatronics: happy to rename it if you have a suggestion!

Funkatronics: renamed to `updateParentId`

PR #1012, opened 2024-10-25
Fix: multi-account bug, same account used across multiple authorizations 

Funkatronics:


android/walletlib/src/main/java/com/solana/mobilewalletadapter/walletlib/authorization/AuthRepositoryImpl.java
```diff
@@ -403,11 +403,11 @@ public AuthRecord issue(@NonNull String name,
         final List<AccountRecord> accountRecords = new ArrayList<>();
         for (AuthorizedAccount account: accounts) {
 
-            final AccountRecord accountRecordQueried = mAccountsDao.query(account.publicKey);
+            final AccountRecord accountRecordQueried = mAccountsDao.query(authRecordId, account.publicKey);
```
Funkatronics, line 406: This lookup is unnecessary now, i think. This check will always return null, and we will create new account entries for the auth record.

Issue #1082, opened 2025-01-27
MWA Not Functioning in WebViews/In App Browsers


Funkatronics:
Local MWA connections from web dapps on Android will not always work if the app is being viewed in an in-app browser or webview. The MWA connection with a local mobile wallet app is imitated by launching a generic URI to open the wallet app and establish the MWA session with the dapp.  These types of links (as well as deep links) are caught by Android Webview by default and not passed to the system to launch the app. Apps that are providing the in app browsing experience must choose to handle these types of links and launch an associated Android Intent if they want to enable this experience for their in-ap browsing context. 

### Symptom
Opening a Solana Web Dapp in an in-app browser on Android (within Telegram, for example) and selecting "Mobile Wallet Adapter" from the wallet connection UI will route the user to the [Solana Mobile Wallets](https://solanamobile.com/wallets) webpage. The user expects their wallet app to be opened when they choose the Mobile Wallet Adapter option, but they are instead taken to the [Solana Mobile Wallets](https://solanamobile.com/wallets) where they are instructed to install an SMS compatible mobile wallet app. 

### Why It Happens
The `wallet-adapter-mobile` js library takes an optional `onWalletNotFound` handler that is called when the MWA intent fails to launch. The default `onWalletNotFound` handler implementation that is provided by the library assumes that the intent failed to launch because the user does not have a compatible mobile wallet app installed and routes users to the [Solana Mobile Wallets](https://solanamobile.com/wallets) webpage so they can install a compatible wallet app like Phantom or Solflare.  

### For Users
Users should open the web dapp in their mobile browsing app (Chrome, Edge, etc) on the Android device instead of trying to access the device from a non-browser apps in app browsing context. Local MWA connections will work on Android when the web app is run in a proper browser app like Chrome. 

### For Dapp Developers
Dapp developers cannot override this behavior to enable MWA connections in these contexts, but they can catch this error and provide and inform the user why MWA cannot be used. The UX here is up to the dapp, but we recommend displaying some kind of error to the user and instructing them to open the website in their browser app instead if they want to use MWA. 

[...]

PR #1083, opened 2025-01-27
Add WebView error handling to Example Web App

Funkatronics:
Adds an error snackbar if a compatible MWA wallet is not found, and the app is running in a webview. See #1082

### DIFF
--- a/examples/example-web-app/pages/_app.tsx
+++ b/examples/example-web-app/pages/_app.tsx
@@ -44,7 +44,7 @@ function App({ children }: { children: ReactNode }) {
                        * Note that you don't have to include the SolanaMobileWalletAdapter here;
                        * It will be added automatically when this app is running in a compatible mobile context.
                        * 
-                       * However, it is recommended to manually include the SolanaSolanaMobileWalletAdapter 
+                       * However, it is recommended to manually include the SolanaMobileWalletAdapter 
                        * here so that you can provide Dapp identity parameters and a walletNotFoundHandler for 
                        * the best UX. The adapter will overwrite any existing SolanaMobileWalletAdapter. 
                        */
--- a/examples/example-web-app/pages/_app.tsx
+++ b/examples/example-web-app/pages/_app.tsx
@@ -12,25 +12,14 @@ import { WalletModalProvider } from '@solana/wallet-adapter-react-ui';
 import { createTheme } from '@mui/material';
 import { ReactNode, useCallback, useMemo } from 'react';
 
-import { registerMwa } from '@solana-mobile/wallet-standard-mobile';
-import { createDefaultAddressSelector, createDefaultAuthorizationResultCache, createDefaultWalletNotFoundHandler } from '@solana-mobile/wallet-standard-mobile';
+import { SolanaMobileWalletAdapter } from '@solana-mobile/wallet-adapter-mobile';
+import { createDefaultAddressSelector, createDefaultAuthorizationResultCache } from '@solana-mobile/wallet-adapter-mobile';
 
 function getUriForAppIdentity() {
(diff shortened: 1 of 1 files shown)


examples/example-web-app/pages/_app.tsx
```diff
@@ -54,7 +43,30 @@ function App({ children }: { children: ReactNode }) {
                       /**
                        * Note that you don't have to include the SolanaMobileWalletAdapter here;
                        * It will be added automatically when this app is running in a compatible mobile context.
+                       * 
+                       * However, it is recommended to manually include the SolanaSolanaMobileWalletAdapter
```
ankur2136, line 47: nit: typo

Funkatronics: SolanaSolana! nice catch, fixed

examples/example-web-app/pages/_app.tsx
```diff
@@ -54,7 +43,30 @@ function App({ children }: { children: ReactNode }) {
...
+                       * here so that you can provide Dapp identity parameters and a walletNotFoundHandler for 
+                       * the best UX. The adapter will overwrite any existing SolanaMobileWalletAdapter. 
                        */
+                      new SolanaMobileWalletAdapter({
+                        addressSelector: createDefaultAddressSelector(),
+                        appIdentity: {
+                            uri: getUriForAppIdentity(),
+                        },
+                        authorizationResultCache: createDefaultAuthorizationResultCache(),
+                        chain: 'solana:testnet',
```
Michaelsulistio, line 57: This should use `Devnet`, consistent iwith the `CLUSTER` variable?

examples/example-web-app/pages/_app.tsx
```diff
@@ -54,7 +43,30 @@ function App({ children }: { children: ReactNode }) {
...
+                        onWalletNotFound: async (mobileWalletAdapter: SolanaMobileWalletAdapter) => {
+                            if (typeof window !== 'undefined') {
+                                const userAgent = window.navigator.userAgent.toLowerCase();
+                                if (userAgent.includes('wv')) { // Android WebView
```
Michaelsulistio, line 61: Is there a more precise way to check for Android Webview? Wondering if there's ever a case where a non-Android Webview also has "wv" in the userAgent.

Funkatronics: from my research, this is a reliable way to detect Android WebView on Lollipop and higher. 

As far as non-android, that should not happen. since MWA is only loaded on Android. If we are on iOS for example. we wont ever get here. 

Even so, if we get here its because MWA failed. There is not much we can do apart from either routing the user to the wallets page, or showing some kind of error. 

I could change up the logic here to do an android check, and only route to the wallets webpage if we believe we are indeed in a compatible browser and therefore the user _likely_ does not have a wallet installed. Honestly IMO its the assuming the user does not have a wallet part that we should be questioning here.

Funkatronics: I think we should change up the UX here entirely. We should never auto route to the wallets page, that UX sucks. We should show a dialog with something like:

> We did not find a compatible wallet installed on the device

And depending on what browser we think we are in either:
> You can see a list of compatible wallets [here](solanamobile.com/wallets)

or:
> This browser does not appear to support MWA

Something like this. Because we don't actually ever know for sure if the user does not have a wallet installed. Its possible they just dismissed the browser exit warning, or dismissed the disambiguation dialog.

Funkatronics: The default implementation routes to the wallets page because there is not much else we can do. We cannot show UI from the library.

Funkatronics:
closing in favor of #1125

PR #1094, opened 2025-02-04
MWA Specification - WebSocket Data Encoding Disambiguation

Funkatronics:
This spec change fixes an existing bug in the Mobile Wallet Adapter WebSocket transport specification. 

The WebSocket protocol defines support for both binary data frames and text data frames. Binary data frames allow endpoints to exchange raw binary payloads while text data frames contain UTF-8 encoded data. Currently the MWA spec does not specify which data frames are supported or how payloads should be encoded for either data frame. 

Currently, our MWA implementations use raw, unencoded binary data frames to exchange the encrypted payloads. Since not all binary payloads are valid UTF-8, the use of text data frames is not valid behavior. This update to the specification disambiguates this behavior for the existing `com.solana.mobilewalletadapter.v1` WebSocket subprotocol and defines a new `com.solana.mobilewalletadapter.v1.base64` subprotocol to allow payloads to be sent as UFT-8 encoded text frames with a shared understanding of how those payloads will be encoded/decoded by the endpoints.


spec/spec.md
```diff
@@ -148,7 +148,12 @@ WebSockets is a mandatory transport protocol for mobile-wallet-adapter implement
...
+1. The `com.solana.mobilewalletadapter.v1.reflector` subprotocol is used by WebSocket reflectors when connecting to a [Remote URI](#remote-uri).
+
+When connecting to a [Local URI](#local-uri), the dapp endpoint must request both the `com.solana.mobilewalletadapter.v1` and `com.solana.mobilewalletadapter.v1.base64` WebSocket subprotocols. The wallet endpoint must respond with one or both of the `com.solana.mobilewalletadapter.v1` and `com.solana.mobilewalletadapter.v1.base64` WebSocket subprotocols to indicate to the dapp endpoint whether it supports either the binary or base64 encoded data frames, or both. When connecting to a [Remote URI](#remote-uri), the both the dapp and wallet endpoints must request, and the reflector server must respond with, the `com.solana.mobilewalletadapter.v1.reflector` WebSocket subprotocol in addition to one or both of the `com.solana.mobilewalletadapter.v1` and `com.solana.mobilewalletadapter.v1.base64` subprotocols.
```
Funkatronics, line 156: @sdlaver I have initially wrote this to allow servers to support both binary and text frames simultaneously, if they choose. Do you think we should make this so that servers should define just one of these, not both?

sdlaver (approved):

J909 (approved):

PR #1166, opened 2025-05-07
Remove web3js usage from wallet-standard-mobile

Funkatronics:
wallet-standard-mobile should not depend on @solana/web3.js. This PR removes dependency on our `mobile-wallet-adapter-protocol-web3js` module as it was unnecessary and created a dependency on web3.js 1.x. `wallet-standard-mobile` now uses `mobile-wallet-adapter-protocol` directly (as it always should have)


Funkatronics:
the name of this branch is misleading. I went in a different direction than originally planned - removing dependency on web3.js instead of migrating to depending on web3.js 2.0/kit

Michaelsulistio (approved):

PR #1225, opened 2025-07-30
[fix] Add HeadlessJsTaskContext to Kotlin module

Michaelsulistio:
# Context

There is an on-going bug with MWA RN library that causes all network requests to hang indefinitely during an MWA session.

This is present in any app that uses MWA with React Native's New Architecture

# Cause
Identified the [root cause](https://github.com/facebook/react-native/issues/50327). 

Summary of issue:
1. In New Architecture, all Timers (e.g `setTimeout`) are paused when the main activity is backgrounded.

2. React Native's `fetch` uses `setTimeout` somewhere in its implementation

3. In MWA, when we `transact` our main activity is backgrounded.

4. In `transact` callback, this causes all network requests that rely on `fetch` (e.g `getLatestBlockhash`) to hang indefinitely because the underlying `setTimeout` never finishes.

# Fix

Initiate a `HeadlessJsTaskContext.startTask()` before we launch the new activity. Starting a headless task will "magically" prevents the main activity host from sleeping, so `fetch`/`setTimeout` will work as expected.

[source](https://github.com/facebook/react-native/issues/50327#issuecomment-3126015895)

## Caveats

- Devs will see a console.warn each time the headless task is started.

``` WARN  No task registered for key MWA_HEADLESS_TASK_KEY```

This is a warning because, the ["proper" API"](https://reactnative.dev/docs/headless-js-android#the-js-api) for a headless task involves registering it on the JS side.

Without deeper investigation, this looks like a harmless warning.

- I noticed some weird behavior where, in a sample app with Expo Router + auto-navigation, I'd get stale UI thread errors after connecting with MWA.

I'm worried there could be some sideffects to the headless task like the ui manager hanging onto dead pointers. 

Need to repro further, but I think we should publish a beta branch to get feedback from blocked devs

### DIFF
--- a/examples/example-react-native-app/yarn.lock
+++ b/examples/example-react-native-app/yarn.lock
@@ -2528,14 +2528,14 @@
     "@sinonjs/commons" "^3.0.0"
 
 "@solana-mobile/mobile-wallet-adapter-protocol-web3js@^2.1.4", "@solana-mobile/mobile-wallet-adapter-protocol-web3js@file:../../js/packages/mobile-wallet-adapter-protocol-web3js":
-  version "2.2.0"
+  version "2.2.2"
   dependencies:
     "@solana-mobile/mobile-wallet-adapter-protocol" "^2.2.0"
     bs58 "^5.0.0"
     js-base64 "^3.7.5"
 
 "@solana-mobile/mobile-wallet-adapter-protocol@^2.2.0", "@solana-mobile/mobile-wallet-adapter-protocol@file:../../js/packages/mobile-wallet-adapter-protocol":
-  version "2.2.0"
+  version "2.2.2"
   dependencies:
     "@solana/wallet-standard" "^1.1.2"
     "@solana/wallet-standard-util" "^1.1.1"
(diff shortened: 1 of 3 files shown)

### CODE REVIEW

File: js/packages/mobile-wallet-adapter-protocol/android/src/main/java/com/solanamobile/mobilewalletadapter/reactnative/SolanaMobileWalletAdapterModule.kt [hunk 1]
```diff
@@ -36,10 +39,12 @@ class SolanaMobileWalletAdapterModule(reactContext: ReactApplicationContext) :
         private const val ASSOCIATION_TIMEOUT_MS = 10000
         private const val CLIENT_TIMEOUT_MS = 90000
         private const val REQUEST_LOCAL_ASSOCIATION = 0
+        private const val MWA_HEADLESS_TASK_KEY = "MWA_HEADLESS_TASK_KEY"
```
Funkatronics, line 42: where is the manifest registry for this task?

PR #1312, opened 2025-09-30
Update bug report template

Michaelsulistio:


.github/ISSUE_TEMPLATE/bug_report.md
```diff
@@ -18,17 +43,10 @@ Steps to reproduce the behavior:
 4. Scroll down to '....'
 5. See error
 
-**Expected behavior**
```
Funkatronics, line 21: I think we should keep this. I added this originally because sometimes people were reporting things and I was unsure what they were expecting to happen. IMO this question helps people think a bit harder about what they are reporting and if its actually a bug.

.github/ISSUE_TEMPLATE/bug_report.md
```diff
@@ -18,17 +43,10 @@ Steps to reproduce the behavior:
...
-A clear and concise description of what you expected to happen.
-
-**Screenshots**
-If applicable, add screenshots or a screen recording to help explain your problem.
-
 **Smartphone (please complete the following information):**
- - Device: [e.g. Saga, Pixel 7, etc.]
+ - Device: [e.g. Seeker, Pixel 7, etc.]
  - Browser: if this issue is related to mobile web [e.g. Chrome, Firefox]
- - Version: if this issue is endemic to specific Android versions [e.g. 22]
```
Funkatronics, line 30: why was android version removed? I guess we never see stuff related to android version but tis certainly possible. especially as we roll out updates and have Seekers on different android versions

.github/ISSUE_TEMPLATE/bug_report.md
```diff
@@ -1,14 +1,39 @@
...
 **Describe the bug**
-A clear and concise description of what the bug is.
+A clear and concise description of what the bug is. 
+
+**Dependency Versions**
+
+A list of relevant dependencies in your app and their versions.
+
+```
+"react-native": "x.y.z"
```
Funkatronics, line 18: what about native consumers? should we have 2 templates, or can we make this more generic?

.github/ISSUE_TEMPLATE/bug_report.md
```diff
@@ -1,14 +1,39 @@
 ---
 name: Bug report
-about: Create a report to help us improve
+about: Create a report with information to help us assist you
 title: ''
-labels: ''
+labels: 'Bug Report'
```
Funkatronics, line 5: this label does not currently exist. did you mean to use the existing `bug` label, or create a new one?

.github/ISSUE_TEMPLATE/bug_report.md
Michaelsulistio: Feels redundant to have a new issue template with this snippet being the only difference. I'll just add another section underneath with the kotlin libs

Funkatronics (approved):

PR #1315, opened 2025-10-02
MWA Specification: Local Reflector Association 

Funkatronics:
Extends the Mobile Wallet Adapter specification to allow a remote WebSocket reflector server to be used for local connections.


Funkatronics:
closing, not planned. Nostr relay covers this case.

PR #1316, opened 2025-10-02
Walletlib Local Reflector Association 

Funkatronics:
implements wallet side of the updated spec in #1315


Funkatronics:
closing, not planned. Nostr relay covers this case.

PR #1378, opened 2026-02-11
Migrate Lerna, Yarn -> Turbo, Pnpm

Funkatronics:
Removes Lerna and Yarn and replaces them with Turbo and pnpm. Many dependencies were updated in this PR as well to get everything working with modern tooling. the ts compilation target was raised from ES6 to ES2022 for compatibility. 

Note: this PR will also resolve nearly all outstanding security alerts for the `js/` directory. 

Testing:
Built all js example apps (RN Dapp, Wallet, and Web App) locally, and ensured that MWA connect + sign flow worked.

### DIFF
--- a/examples/example-react-native-app/metro.config.js
+++ b/examples/example-react-native-app/metro.config.js
@@ -1,5 +1,4 @@
 const {getDefaultConfig, mergeConfig} = require('@react-native/metro-config');
-// const path = require('path');
 
 /**
  * Metro configuration


js/packages/mobile-wallet-adapter-protocol/src/types.ts
```diff
@@ -1,5 +1,5 @@
 import type { TransactionVersion } from '@solana/web3.js';
-import type { SolanaSignInInput } from "@solana/wallet-standard";
+import type { SolanaSignInInput } from "@solana/wallet-standard-features";
```
Funkatronics: no need to import the entire wallet-standard lib. this is more precise and removes some unneeded transitive deps.

.github/workflows/example-rn.yml
```diff
@@ -34,42 +39,30 @@ jobs:
...
-        run: yarn install --frozen-lockfile
+        run: pnpm install --frozen-lockfile
       - name: '[Example Wallet] Type Check'
         working-directory: 'examples/example-react-native-wallet/'
         run: npx tsc --noEmit
         continue-on-error: true # TODO: fix all type issues and remove this line
-      - name: '[Example Wallet] Lint'
-        working-directory: 'examples/example-react-native-wallet/'
-        run: yarn lint
-        continue-on-error: true # TODO: fix all lint issues and remove this line
```
stegaBOB: Assuming we're saving the linting problems for another PR?

PR #1495, opened 2026-05-19
MWA and other stuff transmitted over Relays (Nostr) 

Funkatronics:
Specification for using Nostr relays as an additive transport layer for Mobile Wallet Adapter

PR #1497, opened 2026-05-28
Nostr Relays Implementation

Funkatronics:
Implementation of the [Nostr Relay Spec](#1495)

Big PR, but it can be broken down like a stacked PR:

1. `android/common`: add Nostr association url constants
2. `android/wallet-lib`: wallet endpoint implementation + tests
3. `js/packages/mobile-wallet-adapter-protocol`: web dapp endpoint implementation + tests 
4. `js/packages/wallet-standard-mobile`: wallet-standard wrapper (few changes)
5. `js/packages/wallet-adapter-mobile`: wallet-adapter wrapper (tiny changes)


android/walletlib/src/main/java/com/solana/mobilewalletadapter/walletlib/scenario/NostrRelayScenario.java
```diff
@@ -0,0 +1,338 @@
...
+        synchronized (mLock) {
+            if (mState != State.NOT_STARTED) {
+                throw new IllegalStateException("Scenario has already been started");
+            }
+
+            mState = State.CONNECTING;
+            future = super.startAsync();
+            doTryConnect();
+
+            mConnectionBackoffExecutor = Executors.newScheduledThreadPool(1);
```
oliveeyay, line 116: Think the executor needs to be created before `doTryConnect()`? If `NostrRelay.connect()` throws synchronously it'll hit `onConnectionFailed` -> `doConnectionFailed` -> `mConnectionBackoffExecutor.schedule(...)` while it's still null. Lock's reentrant so I think it's reachable on the first attempt. Maybe move this line above `doTryConnect()`?

android/walletlib/src/main/java/com/solana/mobilewalletadapter/walletlib/scenario/NostrRelayScenario.java
```diff
@@ -0,0 +1,338 @@
...
+
+import java.net.URI;
+import java.net.URISyntaxException;
+import java.util.List;
+import java.util.concurrent.Executors;
+import java.util.concurrent.ScheduledExecutorService;
+import java.util.concurrent.TimeUnit;
+import java.util.concurrent.atomic.AtomicInteger;
+
+public class NostrRelayScenario extends BaseScenario {
```
oliveeyay, line 33: Might want to add some tests for `NostrRelay` / `NostrRelayScenario`? They're the only Nostr pieces without coverage.

android/walletlib/src/main/java/com/solana/mobilewalletadapter/walletlib/transport/nostr/NostrCrypto.java
```diff
@@ -0,0 +1,354 @@
...
+import org.json.JSONException;
+import org.json.JSONObject;
+
+import java.math.BigInteger;
+import java.nio.charset.StandardCharsets;
+import java.security.MessageDigest;
+import java.security.NoSuchAlgorithmException;
+import java.security.SecureRandom;
+
+public class NostrCrypto {
```
oliveeyay, line 27: Could we add the official BIP340 test vectors plus a fixed-key sign-in-JS/verify-in-Java case? Right now we'd only catch interop drift against a real peer.

android/walletlib/src/main/java/com/solana/mobilewalletadapter/walletlib/transport/nostr/NostrCrypto.java
```diff
@@ -0,0 +1,354 @@
...
+    @NonNull
+    public static String computeEventId(@NonNull String pubkey, long createdAt, int kind,
+                                         @NonNull String[][] tags, @NonNull String content) {
+        String serialized = serializeEvent(pubkey, createdAt, kind, tags, content);
+        return bytesToHex(sha256(serialized.getBytes(StandardCharsets.UTF_8)));
+    }
+
+    @NonNull
+    public static byte[] schnorrSign(@NonNull byte[] messageHash, @NonNull byte[] privateKey) {
+        BigInteger k0 = taggedHash("BIP0340/aux", new byte[32]);
```
oliveeyay, line 100: Aux randomness is hardcoded to zeros here. I know that's spec-legal and still produces valid sigs, but we lose the side-channel protection. Could we add a comment noting it's intentional so nobody "fixes" it later?

js/packages/mobile-wallet-adapter-protocol/src/transact.ts
```diff
@@ -637,3 +650,283 @@ export async function startRemoteScenario(config: RemoteWalletAssociationConfig)
...
+                if (type === 'CLOSED') {
+                    disposeSocket();
+                } else if (type === 'EVENT') {
+                    const event = msg[2] as NostrEvent;
+                    if (!event || !verifyNostrEvent(event)) return;
+
+                    switch (state.__type) {
+                        case 'subscribed': {
+                            if (event.content.length !== 0) {
+                                throw new Error('Encountered unexpected message while awaiting reflection');
```
oliveeyay: These `throw`s become unhandled rejections, so the `wallet` promise never rejects and the session just hangs on a bad message. Reject the promise (or ignore + log) instead?

js/packages/mobile-wallet-adapter-protocol/src/transact.ts
```diff
@@ -49,13 +58,17 @@ type JsonResponsePromises<T> = Record<
...
     | { __type: 'connecting'; associationKeypair: AssociationKeypair }
-    | { __type: 'disconnected' }
-    | { __type: 'hello_req_sent'; associationPublicKey: CryptoKey; ecdhPrivateKey: CryptoKey };
+    | { __type: 'connected'; sharedSecret: SharedSecret; sessionProperties: SessionProperties }
+    | { __type: 'hello_req_sent'; associationPublicKey: CryptoKey; ecdhPrivateKey: CryptoKey }
+    | { __type: 'disconnected' };
 
 type RemoteState = State | { __type: 'reflector_id_received'; reflectorId: ArrayBuffer };
 
+type NostrStateNew =
```
oliveeyay: nitpick: `NostrStateNew`, is the `New` leftover? Rename to `NostrState`?

js/packages/wallet-standard-mobile/src/getIsSupported.ts
```diff
@@ -52,6 +52,26 @@ export function getIsPwaLaunchedAsApp() {
...
+        const permission: PermissionStatus = await navigator.permissions.query({
+            name: 'loopback-network' as PermissionName,
+        });
+        return permission.state === 'granted';
+    } catch (e) {
+        if (
+            e instanceof TypeError &&
+            (e.message.includes('loopback-network') || e.message.includes('local-network-access'))
+        ) {
+            return true;
```
oliveeyay, line 69: Returning `true` when the permission name isn't recognized, intentional? Worried it mis-routes on browsers that don't support the permission API.

js/packages/wallet-standard-mobile/src/initialize.ts
```diff
@@ -42,11 +43,22 @@ export function registerMwa(config: {
...
+            registerWallet(
+                new RemoteSolanaMobileWalletAdapterWallet({
+                    ...config,
+                    remoteHostAuthority: config.remoteHostAuthority!,
+                }),
+            );
+        }
     } else {
         // currently not supported (non-Android mobile device)
+        console.warn(`MWA not registered: device or enviroment not supported`);
```
oliveeyay: nitpick: `enviroment` typo.

android/common/src/main/java/com/solana/mobilewalletadapter/common/AssociationContract.java
```diff
@@ -18,5 +18,9 @@ public class AssociationContract {
     public static final String REMOTE_PARAMETER_REFLECTOR_HOST_AUTHORITY = "reflector"; // type: String
     public static final String REMOTE_PARAMETER_REFLECTOR_ID = "id"; // type: Long
 
+    public static final String NOSTR_PATH_SUFFIX = "nostr";
+    public static final String NOSTR_PARAMETER_RELAY = "relay"; // type: String, repeatable
```
oliveeyay: nitpick: comment says `relay` is "repeatable" but we parse it single-valued.

js/packages/mobile-wallet-adapter-protocol/src/transact.ts
Funkatronics: these throws are definitely caught and handled by the wallets, you can see this for yourself on our example app - if any of these exceptions are thrown they are caught and displayed as a toast

js/packages/mobile-wallet-adapter-protocol/src/transact.ts
Funkatronics: lol yes

js/packages/wallet-standard-mobile/src/getIsSupported.ts
Funkatronics: its intentional. what you've described is exactly the correct behavior - a browser that does not enforce the LNA permission supports local websockets.

android/common/src/main/java/com/solana/mobilewalletadapter/common/AssociationContract.java
Funkatronics: handling multiple relays will be covered in a future PR. this one got too big.

android/walletlib/src/main/java/com/solana/mobilewalletadapter/walletlib/scenario/NostrRelayScenario.java
Funkatronics: im hesitant to change our standing code pattern here (used in both existing scenarios) but I'll investigate

android/walletlib/src/main/java/com/solana/mobilewalletadapter/walletlib/scenario/NostrRelayScenario.java
Funkatronics: we dont test scenarios directly unfortunately - the original MWA code base was never made to be testable so some things cant be unit tested (apart from a couple specific cases that were added after the fact). they are tested via instrumented tests in fakewallet.

beeman (commented):
LGTM, no comments on the implementation. 

I'm happy to work on a integration test in a follow-up PR that starts a ephemeral relay and start with the happy path and a few common failing cases.

beeman (approved):

Funkatronics:
> LGTM, no comments on the implementation.
> 
> I'm happy to work on a integration test in a follow-up PR that starts a ephemeral relay and start with the happy path and a few common failing cases.

Thanks Beeman! there is already significant happy path test coverage on both the android and JS sides but always happy to have more.

android/walletlib/src/main/java/com/solana/mobilewalletadapter/walletlib/scenario/NostrRelayScenario.java
oliveeyay: Interesting. Maybe I'm thinking too much, but if the async loop/connector do use the executor at the next line, then there's potential to run in a NPE/NotInitializedException. Most probably we avoid it because most phones have multiple cores/threads so it takes a bit to actually start an action in another thread.

android/walletlib/src/main/java/com/solana/mobilewalletadapter/walletlib/transport/nostr/NostrCrypto.java
Funkatronics: Yes, adding. Its important to note however, we dont secure anything with these keys. they are using effectively as a UUID for the wallet and dapp to find each others topic on the relay. Messages are signed using these keys, but this is on top of MWA's AES encryption. Essentially Nostr cryptography is used only for Nostr compliance, it does not replace MWAs native encryption or any part of the existing specs security posture.

android/walletlib/src/main/java/com/solana/mobilewalletadapter/walletlib/transport/nostr/NostrCrypto.java
Funkatronics: Added - 16 new android side tests + 5 js tests

android/walletlib/src/main/java/com/solana/mobilewalletadapter/walletlib/scenario/NostrRelayScenario.java
Funkatronics: no you are definitely correct. I was under-thinking it. modified both scenarios

PR #1499, opened 2026-06-02
Add shared protocol encoding helpers

beeman:
Expose a protocol encoding subpath for base64, base58, base64url, and array buffer conversions.

Use the shared helpers in the Kit, web3js, wallet adapter, and wallet standard packages, removing duplicated base64 utilities and obsolete bs58/js-base64 dependencies.

Defer base58, base64, and UTF-8 encoder creation until helper use so prepublish tree-shaking can drop unused protocol entrypoint code.

Update the shared tsdown config to emit optional package subpath entries and add a patch changeset for the affected JS packages.


js/packages/mobile-wallet-adapter-protocol/src/encryptedMessage.ts
```diff
@@ -11,7 +12,7 @@ export async function encryptMessage(plaintext: string, sequenceNumber: number,
     const ciphertext = await crypto.subtle.encrypt(
         getAlgorithmParams(sequenceNumberVector, initializationVector),
         sharedSecret,
-        new TextEncoder().encode(plaintext),
+        utf8ToUint8Array(plaintext),
```
Funkatronics, line 15: 😍

js/tsdown.config.ts
Funkatronics, line 1: what are these changes needed for?

Funkatronics (approved):

js/tsdown.config.ts
beeman: Until now, every package only compiled a single entrypoint: `js/packages/<pkg>/lib/esm/index.js`

With this change we can add optional files we want to compile, the one we opt-in to in `OPTIONAL_ENTRY_NAMES` is `encoding`. 

This will compile the entrypoint `js/packages/mobile-wallet-adapter-protocol/lib/esm/encoding.js` that we refer to in [this export](https://github.com/solana-mobile/mobile-wallet-adapter/pull/1499/changes#diff-4b1c599dcf3090b65273e77eb56a79f1290833ed9c343d27cbb9503342a5c634R32)

PR #1502, opened 2026-06-02
Android Gradle Plugin 9

Funkatronics:
Updates a bunch of statements in the gradle files that are no longer valid in AGP 9. 

The only code changes included are in `fakewallet` to resolve some stricter reified type checking in never kotlin versions. 

closes #1402


Funkatronics:
Note: this will not be merged until after the Nostr relay changes have been shipped. 

If possible, I will also wait until AGP 9.0 is supported in IDEA - currently this project cannot be compiled in IDEA Ultimate.

PR #1506, opened 2026-06-02
Refresh React Native examples

beeman:
Rebuild the native projects for both React Native examples from the React Native 0.85 template while preserving the wallet-specific Android entrypoint.

Update the example manifests and lockfiles, replace the deprecated SafeAreaView import, and point walletlib's React Native package entrypoint at its native ESM build with source files available for codegen.

Keep walletlib's promise-based ReactMethod entrypoints Unit-returning for React Native 0.85, preserve missing app identity URIs as null values, and send sign-and-send transactions through the requested Solana chain endpoint.

Remove the stale record-message confirmation waiter from the React Native dApp so successful sends no longer start the unused @solana/transaction-confirmation path; the old AbortController/AbortSignal TODO is gone from that component because the flow now returns the submitted signature directly.

Add a patch changeset for the walletlib packaging fix.


Funkatronics:
Did you build and deploy the apps on an Android device or emulator to verify everything works? I've seen weird runtime errors when upgrading RN in the past

beeman:
@Funkatronics good call! Testing the full flow revealed we needed some changes in SolanaMobileWalletAdapterWalletLibModule.kt plus some tweaks in `example-react-native-app`. Both seem to fully work now.

Funkatronics (approved):

PR #1507, opened 2026-06-03
chore(ci): pin workflow actions

beeman:
Pin GitHub Actions workflow steps to explicit action SHAs with version comments across the repository CI workflows.

Update action versions used by the Android, JavaScript, example app, release, and documentation workflows.

This works well in line with our dependabot config that will now update hash + version tag.


Funkatronics (approved):
nice

PR #1545, opened 2026-06-11
chore(ci): parallelize RN example builds

beeman:
Run the React Native example app and wallet through a matrix job so both example builds can execute concurrently.

Keep the shared project build and example validation steps in each matrix entry.

From ~9 minutes down to ~5. I'll follow up with some more parallelization of gh jobs!

image


Funkatronics (approved):
noooo, i am trying to bankrupt microslop. who's side are you on @beeman? 😜

PR #1546, opened 2026-06-11
chore: group npm dependabot updates

beeman:
Add npm Dependabot groups for common dependency families across the JavaScript workspace and example apps.

Keep Wallet Standard updates together while excluding them from the broader Solana dependency group.


Funkatronics (approved):
ty

Issue #1179, opened 2025-05-29
[Bug] The mobile wallet adapter protocol must be used in a secure context (`https`) in Expo SDK 53


Michaelsulistio:
**Describe the bug**

When using Mobile Wallet Adapter on an Expo app on Expo version 53, users will see this error:

```
The mobile wallet adapter protocol must be used in a secure context (`https`)
```

From this [assertion](https://github.com/solana-mobile/mobile-wallet-adapter/blob/39e1724a17dedc1824040f8b259c553766bc293b/js/packages/mobile-wallet-adapter-protocol/src/transact.ts#L63).

**Solution**

For now, downgrade to Expo SDK 52 while we investigate the root cause and update the MWA libraries.

**To Reproduce**
1. Build and run an Expo app running Expo SDK 53, with Mobile Wallet Adapter RN libraries.
2. Use any MWA method (e.g connect)
3. See assertion error

**Additional context**

Expo SDK 53 included some changes to the `window` object in RN, likely causing that assert to fail now.

c0rdeiro:
did you find any solution for sdk 53?

ignaciosantise:
Adding more info to this:

This issue stems from the fact that the `@solana-mobile/mobile-wallet-adapter-protocol` package is being resolved to the ESM entry point (lib/esm/index.js) instead of the CommonJS native-specific entry (lib/cjs/index.native.js).

In the ESM build, there’s no `.native.js` implementation, and as a result, React Native projects fall back to the web implementation of `transac`, which references `window` and `crypto.subtle` — both unavailable in the React Native environment.

To test this I removed the `lib/esm` folder so React Native resolves to the CJS native entry point, and it works.

PR #1656, opened 2026-09-10
fix: declare the react-native export condition before browser

### LINKED ISSUE #1302: Secure context error in RN
**Describe the bug**
RE-OCCURRENCE of this [issue 1179](https://github.com/solana-mobile/mobile-wallet-adapter/issues/1179#issue-3101448398). Getting an error staying the protocol must me executed in a secure context. 

Image

**Dependencies Version**
```json
 "@solana-mobile/mobile-wallet-adapter-protocol": "^2.2.3",

 "expo": "^53.0.15",
```

**Core Problem**
`@solana-mobile/mobile-wallet-adapter-protocol` package is using ESM build by default instead of Cjs.

Ashutosh0x:
Fixes #1302, and #1179 before it.

Conditional exports are matched in the order the package declares them, not the order the bundler lists its conditions. Metro asserts `browser` alongside `react-native` when it bundles a native app, so with `browser` declared first a React Native app resolves to `index.browser.js` and never reaches `index.native.js`. The browser build still calls `assertSecureContext()`, which is exactly what the `react-native` fork of `transact.ts` exists to avoid, so `transact()` throws ``The mobile wallet adapter protocol must be used in a secure context (`https`)`` on a device where `window.isSecureContext` is not true.

That is why this only started showing up with Expo SDK 53 and React Native 0.79. They enabled package exports by default; before that Metro read the legacy top-level `react-native` field, which was already correct and still is, and it had been quietly covering for the exports map.

`mobile-wallet-adapter-walletlib` is a useful control here. It is the only package that declares no `browser` condition, and it is the only one that already resolved correctly.

This also matters for #1639. That change moved `react-native` and async-storage to optional peers on the reasoning that both are reached only through the `react-native` export condition, which is only true if that condition is the one that actually matches.

The change is key order and nothing else. No target was added, removed or repointed — the manifests are content-identical if you ignore ordering. I checked resolution before and after against an exact implementation of Node's condition matching rather than a bundler, so no defaults get injected: five resolutions move, all of them React Native going from `index.browser.js` to `index.native.js`, and 49 others are untouched, including browser, node, workerd, edge-light and TypeScript. `types` also moves to the front, which TypeScript wants so it matches ahead of the runtime conditions.

`test/packageExportsConditionOrder.test.ts` walks every package that ships an exports map and asserts the ordering plus the resolution a Metro condition set produces. Against `main` as it stands it fails 16 times; with this change it passes. This is the same bug reported twice already, so the test is the part that actually matters.

One thing I left out on purpose: none of these maps declare a `default` condition, so a resolver that matches none of the declared ones resolves nothing at all. That looks like a real second bug, but it changes behaviour rather than ordering, so it seemed wrong to fold into this. Happy to do it separately if you want it.

## Summary by CodeRabbit

- **Bug Fixes**
  - Fixed package resolution for React Native applications so native builds are selected consistently.
  - Prevented secure-context errors that could occur when using wallet functionality in Expo SDK 53 and React Native 0.79.
  - Improved TypeScript package resolution while preserving existing browser, Node.js, and other platform behavior.
  - Added validation to ensure package conditions resolve to the appropriate platform-specific targets.

### DIFF
--- a/js/.changeset/react-native-export-condition-order.md
+++ b/js/.changeset/react-native-export-condition-order.md
@@ -1,0 +1,16 @@
+---
+'@solana-mobile/mobile-wallet-adapter-protocol': patch
+'@solana-mobile/mobile-wallet-adapter-protocol-web3js': patch
+'@solana-mobile/mobile-wallet-adapter-protocol-kit': patch
+'@solana-mobile/mobile-wallet-adapter-walletlib': patch
+'@solana-mobile/wallet-adapter-mobile': patch
+'@solana-mobile/wallet-standard-mobile': patch
+---
+
+Declare the `react-native` export condition before `browser`, and `types` first.
+
+Conditional exports are matched in the order the package declares them, not in the order the bundler lists its conditions. Metro asserts `browser` alongside `react-native` when bundling a native app, so with `browser` declared first a React Native app resolved to `index.browser.js` instead of `index.native.js`. That build keeps the `assertSecureContext()` call the native fork exists to avoid, so `transact()` threw `The mobile wallet adapter protocol must be used in a secure context (https)` on device.
+
+This is why the failure appeared with Expo SDK 53 and React Native 0.79, which enabled package exports by default; before that Metro used the legacy top-level `react-native` field, which was already correct and still is.
+
+Only the key order changed — no target paths were added, removed or repointed. Resolution is unchanged for browser, node, workerd, edge-light and TypeScript consumers; the only resolutions that move are React Native ones, which now reach the native build they were always meant to.
--- a/js/.changeset/react-native-export-condition-order.md
+++ b/js/.changeset/react-native-export-condition-order.md
@@ -13,4 +13,6 @@ Conditional exports are matched in the order the package declares them, not in t
 
 This is why the failure appeared with Expo SDK 53 and React Native 0.79, which enabled package exports by default; before that Metro used the legacy top-level `react-native` field, which was already correct and still is.
 
-Only the key order changed — no target paths were added, removed or repointed. Resolution is unchanged for browser, node, workerd, edge-light and TypeScript consumers; the only resolutions that move are React Native ones, which now reach the native build they were always meant to.
+Only the key order changed — no target paths were added, removed or repointed. The only resolutions that move are React Native ones, which now reach the native build they were always meant to; browser, node, workerd and edge-light are unaffected.
+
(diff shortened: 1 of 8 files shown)


Ashutosh0x:
@Funkatronics @beeman — flagging you two specifically since this sits across both your areas.

@Funkatronics, you were tagged on #1302 a few times. The short version is that the exports maps declare `browser` before `react-native`, so Metro takes the browser build and RN never reaches `__forks__/react-native/transact.ts`. The secure context throw people are seeing is just the unforked `transact.ts` doing its job in the wrong place.

@beeman, this one is relevant to #1639. That PR made `react-native` and async-storage optional peers on the basis that they are reached only through the `react-native` export condition — which is true, but only once that condition is the one that wins. Right now `browser` shadows it in five of the six packages.

Diff is reordering only, no target paths touched. The test walks every package with an exports map and checks both the declared order and what a Metro condition set resolves to, since this is the second time the issue has been filed (#1179, then #1302).

Happy to split it per package, or drop the `types` reorder into its own commit, if either would be easier to review.

Ashutosh0x:
Both CodeRabbit points addressed in b605692.

The subpath one was a real gap — the test only read `exports["."]`, so `./encoding` in `mobile-wallet-adapter-protocol` was being reordered by the fix but never actually covered by the test. It now enumerates every subpath and applies the precedence and Metro-resolution assertions to each. Against `main` that takes it from 16 failures to 19.

On the changeset wording, it was right that I undersold the `types` move. I had it filed under "nothing changes for TypeScript", but types-first is not a no-op — TypeScript matches conditions in declaration order as well, so with `types` last a consumer that also asserts a runtime condition can reach a JavaScript entry before the declaration file. Reworded to say that plainly, while keeping the point that no target path moved.

Ashutosh0x:
Docstring coverage warning is fixed too, in f0d1cfc — all four helpers now carry JSDoc.

Worth recording why `resolveWithConditions` is hand-rolled rather than pulled from a resolver library, since it looks like reinvention otherwise. I originally used `resolve.exports` to measure the blast radius and it reported far more environments changing than actually do, because it injects a `node` condition of its own unless you set `browser`. Asserting exactly the conditions named by the caller and nothing else is what gives the real number: five resolutions move, all React Native, and 49 are untouched.

So the state of this PR now: both review comments addressed in b605692, docstrings in f0d1cfc. Against `main` the test fails 19 times, passes here.

Ashutosh0x:
@coderabbitai review

Both comments from the previous review are addressed: every export subpath is now enumerated in the test (b605692), and the changeset TypeScript wording is corrected. Docstring coverage is handled in f0d1cfc.

Ashutosh0x:
@coderabbitai full review

Funkatronics:
@coderabbitai full review

PR #1657, opened 2026-09-10
fix(walletlib): stop advertising a CommonJS build that is never produced

Ashutosh0x:
`require('@solana-mobile/mobile-wallet-adapter-walletlib')` fails:

```
Error: Cannot find module '.../node_modules/@solana-mobile/mobile-wallet-adapter-walletlib/lib/cjs/index.js'
    code: 'MODULE_NOT_FOUND'
```

The `node` condition declares `"require": "./lib/cjs/index.js"`, but this package builds ESM only. `tsdown.config.ts` emits `format: 'esm'` into `lib/esm`, and `postbuild` writes only the `lib/esm` type marker — compare the other packages, whose `postbuild` writes both `lib/cjs/package.json` and `lib/esm/package.json`. Nothing has produced `lib/cjs` since the build moved to tsdown, so the target has simply never existed.

I checked it two ways. The published 1.4.5 tarball contains `lib/esm` and `lib/types` and no cjs output whatsoever, and a local `turbo run build` emits the same two directories. I also walked the exports map of all six JS packages against their published tarballs — this is the only phantom target in the workspace, the other five are clean.

This points the `node` condition at the ESM build that actually ships. Importers are unaffected. `require` callers on Node 22.12+ load it through `require(esm)`, and on older versions get a plain "cannot require an ES module" error rather than a missing-file one.

If you would rather keep a real CJS entry, the alternative is adding a cjs format to this package's `tsdown.config.ts` and restoring the `lib/cjs` marker in `postbuild` — happy to do that version instead, just say which you prefer.

The test asserts every target in the exports map exists on disk after a build, and that the manifest does not reference a cjs directory this package does not emit. CI builds before it tests so the on-disk checks run there; they skip locally until `pnpm build` has been run.

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed package loading for Node.js consumers using `require()`.
  * Updated package exports to point to the available ESM build, preventing missing-module errors.
  * Improved compatibility for supported Node.js versions while preserving clear errors on versions that cannot load ESM through `require()`.

* **Tests**
  * Added validation to ensure all published export targets exist and do not reference unavailable CommonJS builds.

### DIFF
--- a/js/.changeset/walletlib-drop-phantom-cjs-export.md
+++ b/js/.changeset/walletlib-drop-phantom-cjs-export.md
@@ -1,0 +1,9 @@
+---
+'@solana-mobile/mobile-wallet-adapter-walletlib': patch
+---
+
+Stop advertising a CommonJS build that is never produced.
+
+The `node` condition declared `"require": "./lib/cjs/index.js"`, but this package builds ESM only — `tsdown.config.ts` emits `format: 'esm'` into `lib/esm`, and `postbuild` writes only the `lib/esm` type marker. Nothing has produced `lib/cjs` since the build moved to tsdown, so the file was absent from the published tarball and `require('@solana-mobile/mobile-wallet-adapter-walletlib')` failed with `MODULE_NOT_FOUND` naming a path that had never existed.
+
+The `node` condition now points at the ESM build that is actually shipped. Importers are unaffected; `require` callers on Node 22.12 and later load it through `require(esm)`, and on older versions get a plain "cannot require an ES module" error instead of a missing-file one.
--- a/js/packages/mobile-wallet-adapter-walletlib/package.json
+++ b/js/packages/mobile-wallet-adapter-walletlib/package.json
@@ -6,10 +6,7 @@
     "repository": "https://github.com/solana-mobile/mobile-wallet-adapter",
     "license": "Apache-2.0",
     "exports": {
-        "node": {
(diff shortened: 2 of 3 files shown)


Ashutosh0x:
@Funkatronics @beeman — this one is unrelated to #1656, just found it while auditing the exports maps.

Quick repro if you want to see it, no clone needed:

```
npm install @solana-mobile/mobile-wallet-adapter-walletlib
node -e "require.resolve('@solana-mobile/mobile-wallet-adapter-walletlib')"
```

That throws MODULE_NOT_FOUND on `lib/cjs/index.js`. `import.meta.resolve` on the same install works fine, which narrows it to the `require` branch of the `node` condition.

@beeman, my guess is this dates to the tsdown migration in #1418 — the local `tsdown.config.ts` here only emits esm, and the `postbuild` script only writes the `lib/esm` marker, whereas the packages on the shared root config write both markers. The stale `tsconfig.cjs.json` still sitting in this package looks like it is from the same era. I left that file alone rather than widen the diff, but happy to remove it if it is dead to you too.

I checked all six packages against their published tarballs and this is the only phantom target, so it looks contained.

Ashutosh0x:
Docstring coverage warning fixed in fbf338a — `collectTargets` now carries JSDoc explaining why the whole exports tree gets walked rather than just its top level.

